Issue #78, opened 2015-09-25
DB error when trying to save several services for same date.


tbar0970:
Steps to reproduce:
1. Go to services > Service program
2.  Edit the service program for some congregations, for a date range which doesn't have any services saved yet
3.  The table has editable date fields down the left.  Adjust one of these dates so it's the same as another row.  Enter some service details for both rows.  
4.  Save.
5.  DB error results:  Duplicate value for "datecong".

Expected behaviour:  Before form submission, Jethro checks for duplicate dates and blocks submission.  (Coudl also add better server-side detection, but client-side detection will make sure all the rest of the data in the form is not lost).

edwhiting:
Do you want to make use of MySQL's ON DUPLICATE KEY UPDATE feature at the database level so that it doesn't throw an exception when this happens? http://dev.mysql.com/doc/refman/5.7/en/insert-on-duplicate.html
This wouldn't preclude adding some code on the client side to warn the user and possibly prevent the user from saving the form.

tbar0970:
A silent overwrite isn't the behaviour we want, so ON DUPLICATE KEY UPDATE isn't the solution we need.

To catch it server side, we could override db_object::readyToCreate() in the service class, and add a query that checks for existing services for that congregation and date.  BUT that would necessitate an extra query for each service saved (sometimes there are a lot saved at once).  So given that the existing behaviour is blocking the save correctly (albeit noisily) I think just adding a client-side check would be the best solution right now.

PR #113, opened 2016-01-13
Added ON DUPLICATE KEY UPDATE so that errors don't happen when saving services with the same key.

edwhiting:
This change stopped the errors in issue #78 (https://github.com/tbar0970/jethro-pmm/issues/78) from happening however I haven't tested the change thoroughly; it could break things because it is in a commonly-used function.  It causes the subsequent records in a set of new records to be inserted to clobber any existing records having the same primary key, even records that were inserted in the same transaction.  In the case of issue 78 two services for the same congregation on the same date will have the same primary key.  The first will get lost while the second will be saved.

### DIFF
--- a/include/db_object.class.php
+++ b/include/db_object.class.php
@@ -215,9 +215,13 @@ class db_object
 			array_unshift($flds, 'id');
 			array_unshift($vals, $db->quote((int)$this->id));
 		}
-
+		$updateVals = Array();
+		for($i = 0; $i < count($flds); $i++){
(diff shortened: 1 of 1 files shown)


tbar0970:
I don't think it's a good idea to add ON DUPLICATE KEY UPDATE here.  Silently overwriting existing data is not really the behaviour we want.  And, if it does an update, the call to $db->lastInsertId() a few lines later won't return the right thing, so the newly created object won't have a (correct) id.

PR #119, opened 2016-01-31
Correct footer colspan on Attendance view

howardgrigg:
You'll likely want to check this as I don't know the in's and out of Jethro yet. This works for our install as the footer columns were off by one in the attendance view.


tbar0970:
Perfect.  Thanks!

PR #120, opened 2016-02-01
fix sms gateway issues

vanoudt:
Hello!

Wanted to start using SMS messaging - but noticed that 5centsms (the suggested provider in the conf sample) now has a new API. This allows for using it. Tested, and it works.

This also saves the SMS sent as a note into the person corresponding to the phone number. That way, there's a trail of what was sent etc. Might at some stage look at getting the API for replies working so that that can be automatically added in too - but not in the near term!

### DIFF
--- a/conf.php.sample.au
+++ b/conf.php.sample.au
@@ -180,14 +180,15 @@ define('CCLI_SEARCH_URL', 'http://us.search.ccli.com/search/results?SearchTerm=_
 // Otherwise you can use songsearch:  http://us.search.ccli.com/songs/__NUMBER__
 define('CCLI_DETAIL_URL', 'https://au.songselect.com/songs/__NUMBER__');
 
+////////////////////////////////////////////////////////////////
 // SMS GATEWAY:
-// ------------
+////////////////////////////////////////////////////////////////
 
-// 5centSMS API Token
-define('FIVECENT_SMS_API_KEY', '');
+// The maximum SMS length users are allowed to send.  160 chars is usually a one-part SMS.
+define('SMS_MAX_LENGTH', 320);
 
-// 5centSMS User Email
-define('FIVECENT_SMS_EMAIL', '');
+// The two digit country code for mobile phones. Some providers, such as smsbroadcast, use this
+define('SMS_COUNTRY_CODE', 61);
 
 // Header template for SMS server api authentication.
 // For example, for 5centsms.com.au something like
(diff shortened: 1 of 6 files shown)


tbar0970:
Hi vanoudt

Thanks for the pull request.  

Having a record of what was sent is indeed useful, but given that some churches send everyone an SMS every week, it could easily clog up their notes history quite quickly.  So that feature would need to be disable-able at least.  However I can't actually see any code that implements this anyway?

I'll have a play with the 5cent API changes.  I'd like to see if we can find a more generic way of adding the authentication HTTP headers.

Note to self: Docs at https://www.5centsms.com.au/dashboard/apiDocs#!/v4

vanoudt:
Wow - it appears that I didn't finish committing everything to git :-) My live installation definitely works adding notes when I send an SMS :-) 

I suppose what could be done to make it more generic would be to define: "SMS_AUTHENTICATION_HEADERS" that can be configured in the conf.php. This way, it can be generic, much as the POST URL is.

If I get a chance, I'll have a play with another option to "automatically add sms messages as notes". Perhaps even a checkbox on each SMS modal dialog, which defaults to the configured default! :-)

In terms of the 5cent SMS webhooks for retrieving replies.... I'm not a fan. Mostly because their webhook basically just posts to a page, which we then would blindly put into the system as an SMS reply/note. 5centSMS also doesn't seem to provide any sort of authentication from our side to confirm that they are indeed who are sending the post to us, meaning that it would be very open to abuse (anyone wiht the URL would be able to insert data into the database.) A terrible API!

I'll have a play with this this evening perhaps and see what I can post. (Away from my pastoring desk for the day, so don't have my laptop with me!)

There is another bug, which I should raise an issue for: SMS buttons in reports reloads the page as soon as the dialog shows. I think this might be because the buttons themselves are in a form...

vanoudt:
Hi again,

I've made some changes, which should make things "just work" (TM). Authentication headers (as used by 5centsms can be configured in a very generic way. Also, notes can be enabled/disabled on a per-message instance - with a default set in the conf.php

I've also modified things a bit to work with https://www.smsbroadcast.com.au/ (who are cheaper!). One of the things they do in their api result is to present the number in internationalised form - so that is checked as well. Various sample configuration strings are provided in conf.sample.au for these two providers.

tbar0970:
is this PR made redundant by pull request #179 ?

vanoudt:
It certainly is... sorry, should have closed this one down too. (I have just updated the sample conf files in request #179)

PR #136, opened 2016-02-17
Angular

edwhiting:
I have added Angular 1.5 to the login form.  This is a first step towards separating user interface client-side code from data-generating server-side code.  The aim is to move towards making the code base more modular and therefore more maintainable and agile.

Currently this code sits in the anglr branch, in case you dislike the approach and don't wish to pollute the master branch.


edwhiting:
This branch is no longer relevant: Angular has changed and this branch attempts to impose an architecture on the project that would be too drastic to be worth the while.

PR #143, opened 2016-02-23
Basic character counting

vanoudt:
Adds a simple character counter so that people know how many characters they still have in their sms.
This pull request actually has the right commits in it :-)


vanoudt:
Most of this work is also included (and even more complete) in #141 - if you're happy to do so, I'd rather close this PR and concentrate on #141...

vanoudt:
This is now included in pull request #179, since this one has conflicts, please consider that pull-request instead...

PR #149, opened 2016-02-29
Sort notes descending instead of ascending

vanoudt:
This way, the latest notes are at the top. Also, uses time instead of the string to sort, meaning it sorts with a better granularity. Also, it also considers the comments in the sort. That way, whatever was done last is at the top.

### DIFF
--- a/db_objects/person.class.php
+++ b/db_objects/person.class.php
@@ -311,8 +311,17 @@ class Person extends DB_Object
 
 	function _compareCreatedDates($a, $b)
 	{
-		return $a['created'] > $b['created'];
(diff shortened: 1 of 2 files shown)


vanoudt:
Re-opened after having removed the left-overs from another branch :-)

vanoudt:
This branch can still be applied, but it is also included in pull #179

tbar0970:
Done in a configurable way in https://github.com/tbar0970/jethro-pmm/commit/074443c0eaa3b300200ae8c26efb6c965d164af2

Issue #153, opened 2016-03-02
Archived group members show/hide option back to front


agoddard77:
It seems like by default archived members are shown (greyed out), then when you click 'Show archived members' they are hidden, then when you click 'hide archived members' they're shown again. problem seems the same on demo version.

vanoudt:
Fixed in pull request #156

PR #156, opened 2016-03-03
Fix Issue 153 - include/exclude archived families

vanoudt:
This fix does the trick.
It seems the _family_ db class has a different default for status than the _person_. Both are adjusted to the same coding style for completeness


tbar0970:
Hi vanoudt, thanks for the pull request, but I've fixed Issue #153 in a simpler way.
A parameter like this
```'!status' => '' ```
means "the status is not empty", which doesn't really make sense.

vanoudt:
Very happy to have it fixed in a simpler way! :-)

Issue #139, opened 2016-02-22
SMS link in reports doesn't work


tbar0970:
## Steps to reproduce:
Set up system with SMS gateway
Run a person report that includes mobile numbers
Mobile numbers in results show the phone and envelope icons - hmm.
Click on the envelope icon - page reloads to <base url>/undefined.

## Comments:
It's probably better just to list unadorned mobile numbers in report results ?  Reports are rarely used on phones, so the tel link is not very useful, and to send SMS you can use the bulk action at the bottom.  The tel and envelope buttons are intended just for the view-single-person page.

vanoudt:
Personally, I find the ability to send non-bulk SMSes from reports very useful - use it every week like this. I have a report of those who weren't at church, and send them a "how are you going?" sms. I like to tailor this to the different persons.

This bug has been irritating me for a while, so finally got around to fixing it. The big issue was that the cells were clickable - pull request #141 adds the ability to disable cell-clickability for the mobile_tel column.

As for phone numbers - yeah, probably not much use on computers. Except for those who have skype/some other VOIP plugin installed. Personally - I'd leave it in...

tbar0970:
Fixed in @vanoudt 's SMS-AJAX rework.

Issue #107, opened 2015-12-10
Ability to convert mobile numbers to international format when talking to SMS gateway


tbar0970:
In the conf file you can configure templates for talking to SMS gateways.
Some gateways require the numbers to be in international format (eg +61417123456 instead of 0417123456).
It would be valuable to have some way of making this possible in the configuration.
Perhaps some sort of template for number conversion
```
    define('MOBILE_LOCAL_PREFIX', '0') // remove this from the front of the number when converting to international
    define('MOBILE_INTERNATIONAL_PREFIX', '+61') // add add this to the front of the number when converting to international
````

vanoudt:
Some work towards this in my pull request #120

tbar0970:
This is resolved in aa0f000860008e56bb024c09da28e4f4c100faae based on @vanoudt 's pull request

Issue #57, opened 2015-06-12
FR: Live character count of SMS box content


ken-h:
It would be handy to see how many characters have been used when typing an SMS - or alternatively how many characters remain to be used.

This is really just a "nice to have" feature.

tbar0970:
This is resolved @vanoudt's work which I merged in aa0f000860008e56bb024c09da28e4f4c100faae

PR #179, opened 2016-05-06
Issue 139 rebased (Rework SMS sending)

vanoudt:
Totally rebased SMS sending, since there were some major changes. The old branch doesn't apply properly anymore, this one does, and adds some extra features...


tbar0970:
Thanks for this, and apologies I still haven't got to review your last pull request.  I've been snowed under with other things, but I'm very keen to incorporate contributions and will get to it as soon as possible.

tbar0970:
Hi Nicholas

Thanks for all your work here.  

There are a whole heap of different issues addressed in this pull request.  In future could we please have one pull request per issue because filtering out the different changes took a lot of time.

I've reviewed, improved and merged in the following:
* Character count (issue #57)
* Support for 5centsms api v4
* Support for internationalized numbers in SMS gateway request/response via special keywords (issue #107)
* Support for "array format" recipients in HTTP request (issue #195)
* Ability to save sent messages as notes

I will merge the following items shortly:
* Fixes for sending individual SMSes from reports (issue #139, #205)

I haven't merged the following, which I think need rethinking.  Can you submit separate pull requests if you want to pursue these:
* AJAX changes.  You had run the ajax/json requests through the normal Jethro "view" system but Jethro has a separate "call" system that should be used for ajax/json calls - see the calls/ folder.  Remind me what the benefit of ajax sending is?
* Changes to note sorting.  The desirable order of dates is debatable, but there is no need for special comparison functions - the mysql-format YYYY-MM-DD HH:MM:SS format will sort correctly as text.

vanoudt:
Hi Tom,

Thanks for that - agree completely that this particular pull request got WAY too big! Thanks for taking the time to wade through it! :-)

I will certainly submit new pull requests for those two things. I can definitely rework the AJAX sending to use the "call" system for that.
The benefits of ajax sending are:
* Ability to send SMS without having to reload the whole page (useful for notes etc.)
* Ability to send multiple SMS's (from a report, for instance), and provide feedback as to which numbers sent and which didn't (a new table is provided)
* This is the big one for me - from a report, I have 20 people I want to SMS _individually_. Without AJAX, I would have to reload the page each time I send an SMS. Using AJAX speeds it up considerably(plus, it shows in a new column which people I've already sent to!)

In terms of the notes sorting, I'd be happy to have this as an option. For me, it makes sense to have most recent activity (including comments) at the top.

tbar0970:
Sounds good. Thanks!

tbar0970:
So for AJAX I think what we need is a new calls/call_send_sms.class.php which calls some of the functions I've moved into include/sms_sender.class.php and returns json result.  (Possibly might also need to move getRecipients() out of view_send_sms into sms_sender).

vanoudt:
I'd actually created an include/sms_sender.class.php independently of you! :-) In my version, I had already shifted getRecipients out.
I've merged our two together, putting it in the #213 pull request. At some stage over the next few days, I'll get to incorporating the ajaxification in there too! :-)

views/view_0_send_sms_http.class.php
```diff
@@ -1,168 +1,298 @@
...
-			}
+    $opts = Array(
+      'http' => Array(
+        'method'        => 'POST',
+        'content'       => $content,
+        'header'        => $header
+      )
+    );
+    // To work with HTTP Server errors ourselves, override the system error_handler
+    set_error_handler(null);
```
tbar0970: Hey @vanoudt I hit a snag with this line - passing NULL to set_error_handler only works from PHP 5.5 onwards and many standard linux environment still have PHP 5.3.  Can you tell me more about why this line is needed?  What happens if it's not there?

vanoudt: Hi @tbar0970. That's because of using AJAX. If we don't override the jethro error handling, it puts HTML out, which confuses the ajax parser, which just silently fails. You might think that you've successfully sent an SMS - but actually, nothing has happened. 

To be fair, this really only happens if theres a configuration error. For instance, I can cause it by simply chaning my SMS providers URL to be .c3m instead of .com. But it's still nicer to have some feedback about that, I think!

I've fixed this in PR #213 today - basically, instead of passing NULL, I just pass a custom error handler that converts errors into exceptions (which are then caught by the code). Thank you stackoverflow! :-)

vanoudt: (I should also add that I'd give up on this PR and switch to PR #213 - it's a lot tidier, and applies to master. (And there are all sorts of kinks ironed out of it!)

tbar0970: Thanks for that.  Yes, this PR is closed but I just thought I'd use this venue for the discussion so there is context.  I'll work on merging #213 soon.

PR #183, opened 2016-05-09
Fix spacing of the navbar...

vanoudt:
on pages like admin, so that it doesn't have a 10px gap anymore. The navbar menu etc. there have a 10px gap. Fixed by moving the 10px gap to the #body tag. Seems to be okay even at small screen sizes...

### DIFF
--- a/resources/less/jethro.less.php
+++ b/resources/less/jethro.less.php
@@ -173,7 +173,7 @@ body {
 	box-shadow: 1px 1px 6px rgba(0,0,0,.2);
 }
 #body {
-	margin: 0px 0px;
+	margin: 0px 10px;
 	max-width: 100%;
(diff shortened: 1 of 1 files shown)


tbar0970:
Hi Nicholas, would you mind posting a screenshot of the problem this patch is solving? The flex of the total screen width has been tricky...

vanoudt:
Sure can. Here's what it looks like now:

screen shot 2016-05-09 at 18 16 31

And what it looks like after the fix:

screen shot 2016-05-09 at 18 16 41

vanoudt:
(Of course, I haven't actually changed the colours, that's the less at work.) Might be an idea to automatically cache the less, and not make it a manual job, though...

PR #197, opened 2016-06-03
Fix #195: Add a new SMS template variable for recipients array

### LINKED ISSUE #195: Set and SMS numbers as array
I am using Jethro with SMS Gateway Me (https://smsgateway.me/sms-api-documentation/messages/send-message-to-many-numbers) to send SMS for roaster reminder. 

I have trouble sending sms to multiple users because the _RECIPIENTS_COMMAS_ placeholder would not work with SMS Gateway Me API. Instead, they would need something like : 

`number[]=00000000&number[]=111111111...`

It would help to have another placeholder like _RECIPIENTS_COMMAS_ and _RECIPIENTS_NEWLINES_ with an array : _RECIPIENTS_ARRAY_

mikaoelitiana:
Fixes #195 

Signed-off-by: Mika Andrianarijaona <mika@tag-ip.com>

### DIFF
--- a/conf.php.sample.au
+++ b/conf.php.sample.au
@@ -190,9 +190,13 @@ define('CCLI_DETAIL_URL', 'https://au.songselect.com/songs/__NUMBER__');
 // URL of the server to send SMSes through
 define('SMS_HTTP_URL', ''); // eg http://www.5centsms.com.au/api/send.php
 
+// SMS parameter name for recipients array. Use 'number' to have number[]=...&number[]=...
+define('SMS_RECIPIENT_PARAMETER', 'number');
+
 // The format for the send-sms POST request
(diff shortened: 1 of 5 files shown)


vanoudt:
I've merged this in with my pull request #179 which completely reworks how SMS messages are sent, including nice AJAX based message sending. The only thing I changed was the name of the parameter. Instead of SMS_RECIPIENT_PARAMETER, I chose to go with SMS_RECIPIENT_ARRAY_PARAMETER, to make it clearer that this parameter is only when one uses the _RECIPIENT_ARRAY_ substitution.

mikaoelitiana:
Yes that makes sense. I hope to see this merged soon.  Thanks

tbar0970:
I've applied this changes to the main branch via vanoudt's PR.

PR #204, opened 2016-06-21
Show special dates on the front page

vanoudt:
Allows showing upcoming special dates on the front page.
Also creates a second row on the front page to show it in.

I realise that a person report could be used to generate this information, but it's something that could be really useful to see at a glance when one first logs in.
Not sure if there should be permissions wrapped around this, so that only certain people can see it?

### DIFF
--- a/db_objects/special_dates.class.php
+++ b/db_objects/special_dates.class.php
@@ -5,7 +5,7 @@ class Special_Dates extends db_object
 	// NB This class only exists for the following SQL
 	// It has no ID
 
-	static function getUpcomingDates($personid, $timeframe='4 weeks')
+	static function getUpcomingDates($personid, $timeframe='1 month')
 	{
 		$end_date = date('Y-m-d', strtotime('+'.$timeframe));
 		$current_year = date('Y');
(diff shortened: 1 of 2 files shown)


tbar0970:
Hi Nick, does this show every custom field value, for every person, on the home page for every user ??

vanoudt:
At the moment, yes. (With the limit that it only shows custom fields that have a date associated with them) Not sure what permissions to wrap it in to change that... As far as I can tell, there are no permissions on custom fields at all. That is, if you view a person, you'd see all of their registered custom fields - this just takes less clicks! Am I missing something?

In the long run, it'd be nice to be able to pick and choose the layout of items on the front-page dynamically.... this doesn't do that, though! :-)

tbar0970:
You're right there's no field-specific permissions.  But there are a few things that would need to be addressed to merge this into the trunk:

* People use custom fields for a bunch of different things and it wouldn't make sense to have every date on the home page.  This could be addressed by having a "show on homepage" checkbox next to date fields on the custom-field-config page (and a corresponding column in the custom-field table)
* The query should use the person view not the _person table.  The view is restricted to persons that the current user has access to.
* The new query doesn't really justify a new db_object class - there is no corresponding table.  It could just be a static method on the custom_field or person classes.

A different, and maybe better, way to achieve this would be to be able to mark reports as "show this report's results on the home page."  Then the report could do all the work of generating the results, and it could be used for other applications too.

vanoudt:
Finally had another look at this. I liked the idea to mark reports as "show on homepage" - PR #389 does exactly that! :-)

PR #206, opened 2016-06-22
fix: redirect logout to homepage

### LINKED ISSUE #189: Undefined view when logging out and back in as a different user
To reproduce:
1) Navigate to a view such as admin > user accounts
2) Log out
3) Log in as a user without access to that page
--> error: undefined view

Cause:
?view=admin__user_accounts is preserved in the URL via the logout/login process.

Remedy:
The logout link should go to the home URL (ie no GET vars)

mikaoelitiana:
fixes #189

### DIFF
--- a/resources/js/tb_lib.js
+++ b/resources/js/tb_lib.js
@@ -337,17 +337,21 @@ $(document).ready(function() {
 
 	$('a[data-method=post]').click(function(event) {
 		var p = $(this).attr('href').split('?');
-		var action = p[0];
+		var action = p[0] + '?';
 		var params = p[1].split('&');
+    console.log(params);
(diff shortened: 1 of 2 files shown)


tbar0970:
Thanks for the contribution.  However the tb_lib library is meant to be generic and so we don't really want to have a special case for the "view" param in there:
```
   if (key === 'view') {
 +        action += '&view=' + value;
 +      }
```

The problem was just that because there was nothing before the ? in the logout link's href, the dynamically-generated POST form was re-using the whole current URL as its target, including the "view" param.  So I've fixed it just by making the URL in main.template.php a bit fuller.

But thanks again for having a go, and I look forward to some more pull requests in the future :)

PR #211, opened 2016-06-24
Minify css 

vanoudt:
Use lessphp to produce a minified css.
Still uses less.js in non-production environments, since it seems to be a bit faster. But the size benefits of using lessphp (+minification) make it a winner, in my opinion!


tbar0970:
Hi Nicholas

Just trying to get my head around this.  The way things stood previously was:
1. Dev systems used less.js
2. Production systems with stock styling used compiled CSS (resources/jethro-2.x.x.css)
3. Production systems with custom styling and no cached CSS used less.js
4. Production systems with a manually cached CSS used that (jethro-2.x.x-custom.css)

Am I right in thinking that your changes affect scenario #3?

tbar0970:
Another thought:  Minified CSS is smaller, but it still needs to be generated every time, right?  So scenario 4 would still be better.  NB the CSS is cached in the browser so it only gets requested once per user.

vanoudt:
Hello!

My changes (mainly) only change secanrio 4. Scenario 3 still uses the existing less.js. So, if they don't manually save the custom.css file, it will be generated on the fly (unminified) but otherwise it will be generated not on the fly (minified).
So, 
- Production systems with stock styling use compiled CSS (resources/jethro-2.x.x.css)
- Production systems with custom styling and no cached CSS used less.js
- Production systems with a manually cached CSS use lessphp (coupled with a php minifier)

It'd be possible to switch out less.js (Your scanario 1 and 3) for lessphp, with a simple file, say "jethro.less.live.php" with the following contents:
`<?php                                                                                                            
require_once '../lessphp/lib/Less/Autoloader.php';                                                               
                                                                                                                 
Less_Autoloader::register();                                                                                     
                                                                                                                 
$parser = new Less_Parser();                                                                                     
ob_start();                                                                                                      
include './jethro.less.php';                                                                                     
$less_css = ob_get_clean();                                                                                      
$parser->parse($less_css);                                                                                       
echo $parser->getCss();                                                                                        
?>   `

vanoudt:
Hmmm - thought about it some more. less.js actually takes about 6seconds (on my machine) to compile the less. lessphp is about the same. If we make the scenario use browser caching of 1 day, it should work out well. Last three commits basically get rid of less.js, and do exactly this. So:

1. Dev systems use lessphp, using browser caching (currently set to 1day)
2. Production systems with stock styling used compiled CSS (resources/jethro-2.x.x.css)
3. Production systems with custom styling and no cached CSS use lessphp, using browser caching (currently set to 1day)
4. Production systems with a manually cached CSS used that (jethro-2.x.x-custom.css) [and this is created by lessphp for download, and is also minified]

vanoudt:
For now, this can probably be closed. 3 years on, and it doesn't currently apply. As a work around, I just use custom css, which can be manually minified...

PR #213, opened 2016-07-01
Sms ajax

vanoudt:
~~Not yet ready to be merged, but work in progress~~
This moves sms sending into a helper class, including GetRecipients
~~TODO: setup ajax calling~~
~~TODO in another PR: sms notes~~
~~TODO in another PR: grab fix for sms modals from issue_139_rebased~~


vanoudt:
Almost ready to commit - just need to finish the Roster view :-)

vanoudt:
Hi Tom!
This is now ready for pulling!
:-)
Nick

tbar0970:
I'll merge this next week.  Looks good.

vanoudt:
Thanks Tom!

tbar0970:
Hi Nick

I haven't merged this yet but I've been working through this and polishing a few things.

A few comments and questions:

You shouldn't call a static method on an instance of a class.  Static methods should be called on the class itself.  (It's only because strict mode is turned off that it didn't error).  I've converted SMS_Sender to be all static since it has no member variables.

What was the benefit of SMS_HTTP_RESPONSE_ERROR_REGEX ? Surely if you've got SMS_HTTP_RESPONSE_OK_REGEX defined, and it doesn't find a match for a given recipient, then we can assume it's errored?

Given that $recips is an associative array indexed by personid, there's no need for extra code to remove duplicates.

Default parameters in JS are only supported in Chrome.  

Why was the message box in the send-sms modal an editable div?  Since SMS messages can only be plain text I don't see the need.  

I'm not sure about the way status is shown in the list of persons.
* If I've sent an SMS to one person I'd rather see something right in front of me telling me
* If I've sent an SMS to a bunch of people I'd rather see straight away a list of people who didn't get it so I can contact them in another way - rather than having to scroll through the page inspecting each person's result.
Is there some particular use case that the in-list reporting addresses?

I've also started writing some developer tips at https://github.com/tbar0970/jethro-pmm/blob/master/docs/developer_tips.md

tbar0970:
(Don't commit any changes based on the comments above - I've already addressed them in my local clone of this branch)

vanoudt:
Hi Tom!
Let me see if I can come up with some answers and comments too! :-)

> You shouldn't call a static method on an instance of a class. Static methods should be called on the class itself. (It's only because strict mode is turned off that it didn't error). I've converted SMS_Sender to be all static since it has no member variables.
That makes sense - thanks for picking up on that! I guess I just like things being a bit more object oriented! 

> What was the benefit of SMS_HTTP_RESPONSE_ERROR_REGEX ? Surely if you've got SMS_HTTP_RESPONSE_OK_REGEX defined, and it doesn't find a match for a given recipient, then we can assume it's errored?
Unfortunately that doesn't always work. With our provider (smsbroadcast.com.au) I use the RESPONSE_ERROR_REGEX to see if there is actually an error, as opposed to the system simply not sending through the message. An ERROR response usually means a bad username or password - something that it can be useful to know.

> Given that $recips is an associative array indexed by personid, there's no need for extra  code to remove duplicates.
Hmmm - that's very strange. The reason I added in the code to remove duplicates is because in testing the feature (extensively!) I would often get double-ups sent to my number. 

> Default parameters in JS are only supported in Chrome.
According to https://developer.mozilla.org/en/docs/Web/JavaScript/Reference/Functions/Default_parameters they are also available in firefox... could be worked around, I guess...

> Why was the message box in the send-sms modal an editable div? Since SMS messages can only be plain text I don't see the need.
This was to avoid having to have a _form_ element at all. It seems neater to not put in a form, since we are not actually submitting a form. Seeing as an editable div should be functional in every browser out there today, I'd say an editable div makes good sense...

> I'm not sure about the way status is shown in the list of persons.
> If I've sent an SMS to one person I'd rather see something right in front of me telling me
>  If I've sent an SMS to a bunch of people I'd rather see straight away a list of people who didn't get it so I can contact them in another way - rather than having to scroll through the page inspecting each person's result. Is there some particular use case that the in-list reporting addresses?

As to the case of sending an SMS to one person: If one sends an SMS to a person, the modal dialog should only close automatically upon success. If the message fails to send, it should stay open, and give information about the failure.

As to the case of sending a sms to a bunch of people, the status _is_ displayed in the bulk-sms well - perhaps the well is closing, and not showing it... In essence, you get to see the details all together at the bottom of the page showing which messages failed and why. (In fact, there's a lot of code in the JS to do exactly that!) and _also_ get to see at a glance in the table who failed and who didn't. If, for instance, your next step is to phone someone, you can easily access their record from the table, without having to first find their details.

My particular use case is the following:
Every Monday, I try to send an SMS to those who weren't at church on Sunday. But I like to personalise the messages a bit. I might send the same message to a husband and wife, but a different message to a different family. With the status showing as a column in the table, I get to see at a glance who I have already messaged, and who I still need to message. I guess this might be possible to colour code (that is, highlight rows), but my thinking is in the future such feedback might be useful for other actions (eg. adding to a group, running action plans etc.)

> I've also started writing some developer tips at https://github.com/tbar0970/jethro-pmm/blob/master/docs/developer_tips.md
Excellent! :-)

tbar0970:
> > What was the benefit of SMS_HTTP_RESPONSE_ERROR_REGEX ? Surely if you've got SMS_HTTP_RESPONSE_OK_REGEX defined, and it doesn't find a match for a given recipient, then we can assume it's errored?
> Unfortunately that doesn't always work. With our provider (smsbroadcast.com.au) I use the RESPONSE_ERROR_REGEX to see if there is actually an error, as opposed to the system simply not sending through the message. An ERROR response usually means a bad username or password - something that it can be useful to know. This lets us identify when the API is ERRORing - not just when a message is failing to send.
Ah, I see - it's testing the whole API call rather than an individual message.  I'll move it to SMS_Sender then.

> > Why was the message box in the send-sms modal an editable div? Since SMS messages can only be plain text I don't see the need.
> This was to avoid having to have a form element at all. It seems neater to not put in a form, since we are not actually submitting a form. Seeing as an editable div should be functional in every browser out there today, I'd say an editable div makes good sense... From memory, having form elements caused headaches too...
OK, I just wondered.  It means the character-counting needs to be handled differently.  We could use a textbox without a form.

> My particular use case is the following:
> Every Monday, I try to send an SMS to those who weren't at church on Sunday. But I like to personalise the messages a bit. I might send the same message to a husband and wife, but a different message to a different family. With the status showing as a column in the table, I get to see at a glance who I have already messaged, and who I still need to message.
Thanks, that gives me the idea.

>> Given that $recips is an associative array indexed by personid, there's no need for extra code to remove duplicates.
> Hmmm - that's very strange. The reason I added in the code to remove duplicates is because in testing the feature (extensively!) I would often get double-ups sent to my number. Try having multiple people set with the same phone number. One of the things I found is that, for instance, a family with one mobile number that is shared between two people. Not sensible, perhaps - but code should assume that people will do non-sensical things like that
Ah, yes, two persons with the same number would be the problem scenario.

vanoudt:
Thanks Tom!

Character counting with a div works quite easily too... :-) 😀 I think form elements are designed to be within forms by the spec? Not that that would be the biggest thing...

tbar0970:
> Character counting with a div works quite easily too... :

Yeah, I just wanted to have one handler function rather than a separate handler for the textarea-based version.

tbar0970:
re duplicates: this bit at the top of sendMessage() would catch any lingering dupes:
```
		if (!empty($recips)) {
			foreach ($recips as $recip) {
				$mobile_tels[$recip['mobile_tel']] = 1;
			}
			$mobile_tels = array_keys($mobile_tels);
		}
```

vanoudt:
That makes sense... Could always wrap it in a throwaway form, don't think we'd end up with any nested forms...

vanoudt:
re duplicates, that makes sense :-)

tbar0970:
The thing with adding a column to the table of persons is that the table layout adjusts to make room, which seems to often mess up other things (eg making text wrap where it shouldn't).

Would I be right in thinking that you do your Monday-morning followup using a person report that includes mobile numbers?  How would it be if we turned a person's send-sms button green or red to indicate status of SMS sending, instead of adding a new column?

vanoudt:
I like that idea a lot... Or even changing the whole row colour... 

maybe a class added to the row indicating success or failure, so styling either the sms button or the row could trivially be dome using pure css...

tbar0970:
Yep good idea

tbar0970:
I'll do it now

vanoudt:
:-)

tbar0970:
image

vanoudt:
Looks fantastic!!

tbar0970:
OK, I've committed this with some changes in https://github.com/tbar0970/jethro-pmm/commit/cc5f9dadbd96f86d9076c5b70565c975eadc2709

A few things I changed for architectural reasons:
* Printed the SMS modal in Person::printFieldValue so that it doesn't have to assume some other file has taken care of it
* Instead of passing person ID in to printFieldValue, set the ID on the dummy object in person_query
* Put all the new JS functions into a JethroSMS namespace to keep them together

I made some interface adjustments etc too.  To simplify things for rosters, it just passes the person IDs through rather than getting the SMS library to look up roster assignees which ended up with problems.

Anyhow thanks for your work!

PR #214, opened 2016-07-08
Show mobile phone number in person list

vanoudt:
A very simple PR - just for showing mobiles in person lists.
Ideally, it'd be nice to be able to choose columns to show in the person list - but at least this way one can quickly get a mobile number without having to open the person view.


tbar0970:
Can you change this to only show them on a wide screen - like I did for family phone numbers

vanoudt:
Got distracted for 2 minutes and quickly did that! :-)

vanoudt:
Thanks Tom!

PR #238, opened 2016-08-06
Convert MDB2 to PDO

vanoudt:
Introduces a helper class to switch from MDB2 to PDO. Tested for me, and it seems to work.
Also changes the configuration, to make it a lot easier - writing a DSN is more complex than I think we should expect people to think. Nevertheless, it does attempt to parse the old style DSN from MDB2, and convert it into something for PDO. (Thanks StackOverflow!)


tbar0970:
Thanks Nick, this looks good, I should have time to merge it in next week.

I noticed there are lots and lots of whitespace changes in this PR (some files have *only* whitespace changes).  Can I get you to configure your editor to leave existing whitespace alone so that in future PRs it's easier to see where the real changes are.

Thanks
Tom

vanoudt:
Hi Tom!
Thanks for that. I can remember to do that - I've got git to automatically fix whitespace when I commit changes (for other projects!).

The pedant in me wants to put in a PR just to fix all the whitespace in one go! <grin>

vanoudt:
Decided to finally upgrade today, only to find that my host (siteground) just plain doesn't want to work with the latest Jethro and MDB2. So spent a bit of time re-invigorating this pull request. All rebased off current master, and is working fine for me in production!

tbar0970:
Thanks for this, and many apologies for not merging this PR sooner like I said I would.
I'm really flat out at the moment but I will make 100% sure I get it merged by the end of June at the latest.

vanoudt:
That's quite ok... Hoping and praying your busy months go well... Appreciate all you do...

tbar0970:
Hi Nick

Thanks for your patience.  I've been looking through this and have a few questions:

1) It looks like there are a lot of new calls to  closeCursor().  The manual says "This method is useful for database drivers that do not support executing a PDOStatement object when a previously executed PDOStatement object still has unfetched rows".  

In my experience this kind of scenario would be something like
```
while ($row = $stmt->fetch()) {
   if (some special condition) {
       $stmt->closeCursor();
       return $row;
   }
}
```
ie, when there may be more results to come but we've given up on fetching them.  
But there are barely any instances like this in Jethro.
Is there some other reason for calling closeCursor() that I'm missing?

2) In the JethroDB constructor was there a particular reason for setting the PDO::NULL_EMPTY_STRING option to true?  Normally in jethro we want blank strings to stay as blank strings.

3) For some positive I was thinking we could get rid of all the check_db_result calls, and do such checking within the JethroDB class - does that sound OK?

vanoudt:
Hmmm - good points.
1. Good point. I've pretty much removed all of this.
2. No idea why I did that. Removed it, and everything seems to still work fine!
3. Done this - and moved most of the error checking into JethroDB. Pull request updated

vanoudt:
PS: No worries at all - this is just to scratch an itch of my own! I'm in no rush... my site is working with this with no issues!

garrettcam:
Hi I'm trying to get jethro to install on a fresh Ubuntu 16.04.01

I create MySQL user and db
```
CREATE USER 'newuser'@'localhost' IDENTIFIED BY 'password';
GRANT ALL PRIVILEGES ON * . * TO 'newuser'@'localhost';
CREATE DB jethro;
FLUSH PRIVILEGES;
```
I clone and pull
```
sudo git clone https://github.com/vanoudt/jethro-pmm.git jethro
sudo git pull origin MDB2PDO
```
I edit conf.php. 

When I load jethro for the first time to install. I get the following error. Any help would be much appreciated.

Warning: Declaration of JethroDB::quote($string) should be compatible with PDO::quote($string, $paramtype = NULL) in /var/www/html/jethro/include/jethrodb.php on line 0

Fatal error: Uncaught Error: Call to undefined function db_error() in /var/www/html/jethro/include/jethrodb.php:78 Stack trace: #0 /var/www/html/jethro/include/user_system.class.php(89): JethroDB->queryRow('SELECT count(*)...') #1 /var/www/html/jethro/index.php(64): User_System->hasUsers() #2 {main} thrown in /var/www/html/jethro/include/jethrodb.php on line 78

vanoudt:
Hi Garrett,

Hmmm - that's strange. It works fine in production for me. I've thrown up a quick fix - which still works fine for me, but will hopefully make that error go away for you! Simply repeat all your steps, or just modify the file jethrodb.php in the include directory. Repace:
`  public function quote($string) {
	if ($string === NULL) {
		return 'NULL';
	} else {
		return parent::quote($string);
	}
  }`

with 
`  public function quote($string, $paramtype=NULL) {
	if ($string === NULL) {
		return 'NULL';
	} else {
		return parent::quote($string, $paramtype);
	}
  }`

tbar0970:
@vanoudt I reckon the "override should be compatible" error depends on what level of strictness is set in php.ini

But it sounds like the "Call to undefined function db_error() in jethrodb.php:78 " error is something separate...?

vanoudt:
@tbar0970 Good point!
And I had completely forgotten about that part of the error. A silly mistake - forgetting to prepend self:: to the db_error calls! All fixed now

@garrettcam I suggest you reclone to get the latest - and let me know if it is working for you!

garrettcam:
@vanoudt That solves the first error.
self:: now means db_error() is called

But there was an error in db_error() so I edited the function in jethrodb.php:23

public function db_error($exception) {
    trigger_error($exception->getMessage(), E_USER_ERROR);
    exit();
  }

Now I get

Fatal error: SQLSTATE[42S02]: Base table or view not found: 1146 Table 'jethro.staff_member' doesn't exist 

This I think comes from hasUsers() and is used to determine if we need to install for the first time.

vanoudt:
Hi @garrettcam

Thanks for that. That's what I get for not actually writing proper tests to check what happens if there is an error! Good thing you spotted that!

But the "good" news is that the db_error function is now telling us what the problem is. For some reason we can't seem to find the jethro.staff_member table.... I wonder if it was created properly in the first place? Would it be possible for you to show a list of tables from your database? (From the CLI - `mysql -uUSERNAME -pPASSWORD` then `use JETHRODATABASE;` then `show tables;`

It's been a while since I did a fresh install of Jethro... @tbar0970 do you have any ideas?

vanoudt:
Hang on - I'm onto it! It's because of shifting the error handling into jethrodb. Will fix ASAP!

tbar0970:
The absence of the staff_member table means the system has not yet been installed.  So it shouldn't result in a fatal error, but rather should lead to the launching of the installer.

tbar0970:
it's the one time a DB "error" shouldn't be fatal.

garrettcam:
@tbar0970 yes this is a fresh install and the installer would not start. My fix to get the error is probably not best practice, so please review it properly as haven't looked at php in over 10 years.

garrettcam:
could the detecting of installation be performed by a file check rather than a db table check?

eg

```
<?php

// HAS INSTALLATION OCCURED
if (file_exists( __DIR__ . '/installed.php' )) {
    echo "YES: LOAD LOGIN PAGE";
} else {
    echo "NO: RUN INSTALL";
    // INSTALL STUFF HERE
    touch( __DIR__ . '/installed.php' );
}

?>
```

tbar0970:
@garrettcam a file check is possible but since databases get transferred between servers etc I'd rather look at the actual DB state.  Previously we were able to catch the table-not-found situation without crashing - we just need to find the way to do it in this MDB2-free branch. We'll get there!

vanoudt:
fixed locally - but I've got a couple of sermons to finish off... will update the pull request as soon as I can!

vanoudt:
Add just pushed changes. Tested - install works fine now.

@tbar0970 One question: do we really need different checks for persons and for staff members? There were different SQL checks in index.php and installer.class.php....

tbar0970:
The check in index.php and installer.class.php run at different times.  The one in installer.class.php is just a sanity check to make sure we don't try to install on top of an already-installed system.  In both cases it doesn't matter which table it looks at  - the presence of either of those tables tells us the installer has already run.  
I've review and merge this soon.

vanoudt:
@tbar0970 That makes sense! (Oh - and no rush :smile:

tbar0970:
Merged manually into master with various mods.
Needs more testing in master before release.

vanoudt:
Thanks Tom.
I notice that you've not included any of the error checking in the jethrodb object... but added it in to the global system - great idea! -)

tbar0970:
Yeah, letting most DB exceptions fall through to the top level means we have the opportunity to catch them beforehand if we want to, eg when detecting whether to run the installer.  Simpler than an extra parameter saying "don't worry if there's an error"!

Thanks for all your work porting the MDB2 code - great to have that antiquated library gone.

vanoudt:
That's a pleasure! :-) Thanks for getting it into a merge-able state! (Including comments!)

I've got it running live(!!) and will let you know if any faults pop up!

PR #242, opened 2016-08-08
Php7changes - silence warnings under PHP

vanoudt:
This silences warnings under php7 - it's developed on top of #238, so I'd suggest merging that one first...

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,14 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->exclude('include/phpword')
+    ->exclude('include/swiftmailer')
+    ->exclude('resources')
+    ->in(__DIR__)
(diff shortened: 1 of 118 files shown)


tbar0970:
This PR seems to have all your PDO/MDB2 changes (and whitespace changes!) in it too.  Can you create a PR with just the PHP7 changes?

tbar0970:
Oh now I see your comment at the top.  Hmm, I will come back to it when I've merged the PDO changes.  I'm not a git expert bit I would have thought it will let you include just certain changes in a PR?

vanoudt:
Hi Tom,

Problem was that I branched off my MDB2PDO branch, instead of off master....
If you'd like, I've put in a new PR #244 which is based off master (just cherry picked the commits I think are php7 changes into there...

Nick

PR #245, opened 2016-08-09
Whitespace fixes

vanoudt:
Here's the pedant in me, putting in a PR that does nothing but fix whitespace in the PHP files! :-)
I used the tool for here: http://cs.sensiolabs.org/ to do the job - it's got some really nice features there than can be used to check/fix the files so that the coding standard used complies with PSR-2. Not sure if there's currently a coding standard used? 

Also included in the PR is the config file for the tool - if you choose, one could just select a subset of "fixes" to make sure that new commits etc match? 

The config file included also ensures to not touch the phpword or swiftmailer classes, and also won't touch anything under resources - we don't want to mess with upstream files...

Of course, you could decide to just leave this alone and close this - just thought I'd suggest it <grin>

### DIFF
--- a/.php_cs
+++ b/.php_cs
@@ -1,0 +1,14 @@
+<?php
+
+$finder = Symfony\CS\Finder\DefaultFinder::create()
+    ->exclude('include/phpword')
+    ->exclude('include/swiftmailer')
+    ->exclude('resources')
+    ->in(__DIR__)
+;
+
+return Symfony\CS\Config\Config::create()
+    ->level(Symfony\CS\FixerInterface::NONE_LEVEL)
+    ->fixers(array('trailing_spaces', 'encoding'))
+    ->finder($finder)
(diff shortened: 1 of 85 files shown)


tbar0970:
Sounds great.  I've generally been following the Squiz coding standard that I was familiar with when I first started Jethro - broadly speaking, tabs for indentation, opening braces on their own line for classes and functions but not for control structures, camel case for class methods, underscores for standalone functions.  I've become a bit sloppy on camel case vs underscored variable names, I admit.

I'm happy to adopt one of the PSR standards - which one do you think is closest to the existing Jethro code?

tbar0970:
Just had a look at CS Fixer.  It's got a lot of features!  I'll have a look through and make a list of those I think should apply to Jethro.  But in the meantime I'll merge this PR in soon.

PR #249, opened 2016-08-17
Material ui redesign

vanoudt:
Hi Tom,

I've been playing around with doing a complete UI redesign of Jethro, making it based on bootstrap 4, and Google's material UI. I'm already managing to get it a lot more responsive, without having to do sniffing of isNarrow or things like that. Plus, I think it brings the UI up to a more modern look.

This is _NOT_ for pulling right now - it's at a VERY basic level. But already, there's some signs of it being easily doable..

Some screenshots showing the basic look. Colours etc. need to be adjusted at some point - these are just using basic material design colours at the moment...
phone4 - login
iphone4 - family list
view family list - fullscreen

Benefits of such a rewrite:
- A more modern UI
- Fully responsive, same code regardless of screen size in PHP
- We can move the search into the application search, so only have one place for it
- Bootstrap 2 is getting very old...

What do you think?


tbar0970:
Agree that bootstrap 2 is getting old.  But having integrated bootstrap into Jethro the first time around, I can tell you this is a BIG job.  There are lots of pages that would need adjustments and lots of testing needed.  You're welcome to have a go, but to be honest I'm not sure when I'd have enough time to review it thoroughly enough to merge it into a release.

vanoudt:
Yep - can believe that it's a big job. Fortunately, it seems to be going quite quickly over here - with minimal changes to the code. I'll keep plugging away at it - since it is a nice distraction now and again. :-) As someone who uses Jethro a lot on my phone and tablet, I sometimes miss a totally responsive UI!

Some elements, such as the responsive tables shown above, could very easily be back-ported into bootstrap2 (just some nicer CSS, and an extra data- element on td cells...)

tbar0970:
OK.  I agree there is room for improvement in the small-screen experience particularly, and I don't want to be a bottleneck. 

What I did with the previous bootstrap implementation, and I think we'd need to do again, is make a spreadsheet with all the different pages as rows and target browsers as columns, and tick off what had been adjusted and tested.   Perhaps we could recruit a third person to do this testing.

tbar0970:
BTW the current server-side responsiveness using Size_Detector::isNarrow() etc is only used in a few cases - it's mostly client-side.  THe server side stuff is used when either the markup needs to be completely different (eg accordion vs tabs) or there's a bunch of data we don't want to send across the wire to a small device that can't display it.  I suspect we would want to retain some of this.

maxpdx:
Any updates on this PR?

vanoudt:
Haven't had any time to work on this for some time. Closing it down for now. Would still love to see a UI overhaul for Jethro - to use more modern CSS features like grid and flexbox - but realise that that's a HUGE task!

PR #252, opened 2016-08-25
Fixed locking to allow recording attendances

funkydan2:
There were two lines where the longer 'attendance-' was still in the code.


tbar0970:
Perfect.  Thanks.

PR #260, opened 2016-09-02
Set sql_mode to avoid strict mode errors

funkydan2:
I wonder whether Jethro should set the appropriate MySQL modes per session, rather than relying on the admin needing to know about such things. (It seems that the errors are popping up for a few people.)

This seems to be working for me.


tbar0970:
It's a good idea, but since it adds an extra query to every page load, could we make it a config option?

funkydan2:
That makes sense. Do you think the default should be to not set SQL mode?

tbar0970:
Yeah let's default to off, to avoid the efficiency cost when it's not needed.
It'd be good to add a comment above the conf.php item (which should be added to all the sample confs) saying "turn this on if you get database errors mentioning only_full_group_by".

funkydan2:
Sorry it's taken so long to get to 2 lines of code. I've now set the query to only be run if STRICT_MODE_FIX is set TRUE. By default this setting isn't defined.

PR #262, opened 2016-09-05
Sms fixes

vanoudt:
Hi Tom!

Just a quick fix - there's an argument missing in the function for bulk sms messages - meaning they plumb don't work! This should fix that!

Oh, and a bonus behind-the-scenes spelling fix too! :-)


tbar0970:
I had somebody report this earlier today - perfect timing!
Thanks for the fix.

Issue #185, opened 2016-05-09
'Contact List' view not available to non-admin user


funkydan2:
When I create additional users, the Families->Contact List view doesn't appear available to them. I think that this view should be available for any user who is has permission to view reports and groups.

tbar0970:
Currently requires sysadmin permission, but there's no good reason for this to be the case.

funkydan2:
Thanks Tom!

PR #287, opened 2016-11-05
Allow 'Contact List' for non admin user

funkydan2:
Regression bug. This was fixed a while back in #185.


tbar0970:
Thanks for the fix.  Looks like I let the regression slip in with another pull request a few months back - apologies.

PR #365, opened 2017-05-23
Fixes to SMS error reporting

vanoudt:
Turns out I had a configuration error when I upgraded - and figuring that out took ages! Partly because I wasn't using an ERROR regex, and partly because the ERROR regex functionality just plain wasn't working. This code fixes that, meaning that people with broken configurations should be able to figure it out more easily. Also logs errors to the Javascript console, for those who don't have an ERROR regex set up (which is, admittedly, a more advanced use case!)


vanoudt:
Thanks @tbar0970

PR #376, opened 2017-06-27
Fix SMS error checking

vanoudt:
Used an undefined variable by mistake, didn't wrap the regex with delimeters. D'oh!
This makes sending SMSes work again. Not sure when/how it broke...


vanoudt:
Oops... Forgot about pr #365 ....  that fixes this much more comprehensively!

Issue #123, opened 2016-02-10
SQL erros on new install


funkydan2:
I'm getting a few SQL errors with a fairly new Jethro install.

For example, if I go to Services->Component library, I get the error:
`_doQuery: [Error message: Could not execute statement] [Last executed query: SELECT service_component_category.id, service_component_category.category_name, service_component_category.runsheet_title_format, service_component_category.handout_title_format, service_component_category.is_numbered_default, service_component_category.show_in_handout_default, service_component_category.show_on_slide_default, service_component_category.length_mins_default FROM service_component_category] [Native code: 1054] [Native message: Unknown column 'service_component_category.is_numbered_default' in 'field list']`

So I assume the issue is there's something missing in my database.

The origin of this may be that I originally installed from the current GIT master, but noticed a few issues. So re-read the documentation and down-graded to the latest stable version. Lots of things are working now (adding people, marking attendance) but not everything.

Any ideas of a way forward? (e.g. could I drop the database, let Jethro create a new one, and then import that data I've entered?)

tbar0970:
Hmm, yes it's because you've installed the latest GIT version and then downgraded.

If you drop the DB and re-install, migrating your existing data will be tricky (Jethro can import a CSV of persons and service components, but not other things).

v2.14 will be coming out in the next week or so, so the best approach might be to wait for that, upgrade your code, and then look through the 2.14 SQL upgrades to see which ones you need to apply and which ones you've already got.

funkydan2:
Thanks so much Tom. I think your proposal sounds good.

funkydan2:
Tom - just a massive thanks for your response, and a note to say that the release that came out last night has resolved all the persons. Your work is a great blessing to a minister in a small church in regional Australia.

Do you have a system for donations from people who don't have a need to sponsor new features?

tbar0970:
Great to hear the fix worked and that Jethro is benefiting your ministry!
We don't have a donations system, but I could suggest some small things you might like to sponsor.
Can you drop me an email - my github user name @gmail.com

PR #389, opened 2017-07-26
Front page reports

vanoudt:
Allow reports to be set to show on the main home page. Any reports can have the box ticked to show their results on the main page.
There is also another checkbox to allow overriding the necessity to have the "RunReports" permission. My thinking on this is that some reports should be allowed regardless of this... Perhaps a new permission should rather be added - something like "Allow running reports as widgets"... but I'm not sure how the permissions system works. What do you think @tbar0970 ?


vanoudt:
Hi Tom! Any thoughts on this. Using this in production, and it's really useful. This way, we can see, for instance, upcoming important dates without having to manually go run a report...

tbar0970:
Sorry for the delay!  Just moved house and I'm not sure I'll have time to review and merge it for a few weeks.  But the idea is a good one so I'll merge it eventually!

vanoudt:
Not a problem 😀

tbar0970:
Just had time for a quick look at this.  A few suggestions:
1. What's the reason for having an additional table frontpage_person_query instead of just adding two columns to the person_query table?  Seems to add a lot of complexity that could be avoided.
2. On the homepage the title is "Person Report #123 results" - wouldn't it make more sense to use the saved name of the query?
3. For code cleanliness can we please not have SQL in view_home.class.php.  SQL belongs in db_object classes.  If you followed suggestion #1 above you could just do something like
```
 $GLOBALS['system']->getDBObjectData('person_query', Array('show_on_homepage' => 1))
```

vanoudt:
Have just modified this according to your suggestions.
1. I was using an additional table because it seemed more sensible to separate out details that weren't needed in that table. But instead, I've now used a single column added to the person_query table.

2. That should only have arrived if there wasn't a name associated with the query. This shouldn't occur anymore with the new design

3. Done!

tbar0970:
Merged manually - see 61f04af51a39d70864a38d863daf59291503e1ce

To avoid magic numbers in the database, I changed the new column to a varchar which contains either blank string, 'auth' or 'all'.  You'll need to update this column definition on your system that already has these changes.

I also pulled out the whitespace changes into a separate commit - hopefully this is the end of them!

PR #393, opened 2017-08-01
Roster reminder script and ini file

s4069b:
Here is a first go at this .. improvements welcome.

### DIFF
--- a/scripts/email_report.php
+++ b/scripts/email_report.php
@@ -52,31 +52,30 @@ if (!(int)$ini['REPORT_ID']) {
 $report = $GLOBALS['system']->getDBObject('person_query', (int)$ini['REPORT_ID']);
 $reportname = $report->getValue('name');
 if (empty($reportname)) $reportname = 'Jethro-Report-'.date('Y-m-d_H:i');
+//first grab the csv for the attachment
 ob_start();
 $report->printResults('csv');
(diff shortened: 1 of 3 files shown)


s4069b:
I have had another go at the email_report.php script and the way that it handles the html content of the email. this should avoid the previous issues with miss-handling spaces and newlines.

s4069b:
ah... I actually broke the email_report.php script for reports that do not include the 'check-box' option. $html_string = strstr($html_string, 'With selected persons:', true) returns FALSE in that case. 
so change the line66 to this instead (i.e. so if 'with selected persons' is not found in the string it will instead return the string un-altered). Hope that makes sense. 

$html_string = strstr($html_string, 'With selected persons:', true)?: $html_string;

tbar0970:
It's a good move to let the person_query object generate the HTML for you.  But I fear that stripping stuff out after the fact is not the best approach - too much to miss, to hard to strip exactly the right content.  

Perhaps you could pass a parameter to the printResults function to indicate you want "basic" HTML without checkboxes etc?

s4069b:
I can have a look at that. thanks for the idea

s4069b:
Somehow I've managed to get two script caught up in one pull request
- the roster_remider script
- changes to the email_report script

PR #411, opened 2017-09-21
Smsencoding fixes

vanoudt:
Fix to only allow plain-text sms messages, and also to restrict pasting to GSM0338. Other filters can be added in the future. 

SMS encoding is currently set in the conf file - Tom, what's the process for adding it in to the database-based settings?


vanoudt:
I think Tom fixed this elsewhere? Closing this old PR for now.

Issue #163, opened 2016-03-09
Person reports: Ability to filter date fields before AND after the current date


tbar0970:
Currently you can filter by a date field value being within X days either before OR after the date on which the report is run.
It would be valuable to be able to filter by, for example, from 30 days before til 30 days after the report date.

tbar0970:
Also to filter for dates that are MORE THAN x days before/after the current date.
Eg training dates more than 365 days ago.

vanoudt:
I have a local branch where I am currently replacing all of the date fields in mysql to be DATE fields rather than text fields. Things like this are then very very easy to do using plain SQL.... looking at the SQL produced by reports, using DATE types for dates drastically reduces the complexity of the SQL, which can only make the queries much faster.

For dates that don't have a year associated with them, I simply set a very early leap year as the year, and in php filter that out/in as required...

tbar0970:
Hmm, interesting.  Your queries would still need to have special handling for your "magic year" dates, eg if I'm looking for exact value or anniversary in the last 30 days, I'd need to figure out the relevant period within the "magic year" and filter for that too.  Which gets complex if it's 10th January and I'm searching for dates or anniversaries in the last 30 days.  

Which is to say:  I'm not sure that the existing complexity is caused by the fact that the column isn't of type DATE - I think it's inherent to the situation of having year-less dates in the mix.  MySQL is quite happy to cast values as dates along the way anyway.

I don't think this particular FR would increase the complexity of the queries - it just requires that the interface be extended to allow for more complex offsets.  Making that possible whilst keeping the interface easy to understand will be the hard bit.

But maybe I'm missing something...

vanoudt:
Actually, one can use the MYSQL DAYOFYEAR function to easily do things like
that. The "magic year" is simply one that is a leap year, so DAYOFYEAR
stuff should work ok....

Agreed that the current TEXT type works, and casting is definitely useful,
but being able to use the MYSQL functions (like DAYOFYEAR) makes things a
whole bunch easier... In my report test, the SQL WHERE clause drops to
about a third of the size - easier reading, easier to understand etc. etc.




------------------------------

tbar0970:
I'll be interested to have a look :-)

vanoudt:
Just threw up the current code as a pull request...

Have a look at the person_query.class.php, around line 920. We go from:
`							$qFromYear = $db->quote(substr($from, 0, 4));
[...]

PR #413, opened 2017-10-04
Dates are dates

vanoudt:
Idea to see how change date fields in mysql could possibly simplify and speed up Jethro queries, and possibly make the interface easier.
Also includes some experiments to see about using date pickers in the interface.

NOT ready to be merged, just here to show what I'm working on :-)
Perhaps of interest to things like #163


vanoudt:
Haven't looked at this for quite some time now. Closing it here to avoid noise!

Issue #429, opened 2017-12-28
create_function is deprecated in PHP7.2


tim-pearce:
PHP create_function() appears to be used in the db_objects classes.

In PHP7.2 this causes messages like this to appear in the logs
[Thu Dec 28 14:14:27.424627 2017] [php7:notice] [pid 919] [client 127.0.0.1:46004] Function create_function() is deprecated - Line 50 of /DATA/htdocs/jethro/db_objects/abstract_note.class.php, referer: http://scott/jethro/index.php?view=_edit_person&personid=1
and Jethro displays 
An error occurred. Please contact your system administrator for help.

tim-pearce:
I can fix this, but I've never contributed to a github (or any online) project before so it may take me a while to figure out how to contribute.

tbar0970:
Merged

PR #430, opened 2017-12-28
Replace create_function

tim-pearce:
Replace all uses of create_function with anonymous function as per issue #429


tbar0970:
Thanks for this.  Looks good.

I'd like to make sure Jethro is still compatible with PHP 5.3.  Anonymous functions were introduced in 5.3 exactly so that's cool.  

But a general note for the future: Some aspects of anonymous functions were introduced in PHP 5.4 (http://php.net/manual/en/functions.anonymous.php) so let's not use them at this point.

Issue #441, opened 2018-02-04
Import creates duplicates


tim-pearce:
When you do an import from CSV and the person already exists (or you have a duplicate in the CSV) a new person is created with the same name.
This is quite inconvenient as Jethro does not allow people to be deleted.

tbar0970:
18. Check consecutive rows with identical family details but with extra whitespace after suburb in one row end up in same family.
19. Row n has a mobile number. Row n+1 has blank mobile.  Make sure the mobile number doesn't bleed down.

tbar0970:
Test plan for complete solution:

- [ ] Matching on name only
- [ ] Matching on name + email
- [ ] Matching on name + mobile
- [ ] Matching on name + email + mobile
- [ ] Matching on name where there are 2+ matching existing records - expect warning
- [ ] import family with 3 members where the THIRD one matches an existing person - should create the first 2 under that family.
- [ ] Import family with 3 members where the FIRST one matches an existing person - should create the last 2 under that family
- [ ] Import family with 3 members where the MIDDLE one matches an existing person - should create the surrounding persons under that family.
- [ ] Importing custom fields of all types.
- [ ] Importing invalid custom field value - expect error
- [ ] Unrecognised column warning
- [ ] Check that blank values in import never overwrite non-blank existing values
- [ ] When existing person/family and import data both have different non-blank values, check the "preserve existing" option works.
- [ ] Family member 1 has address, family member 2 has home phone - should end up in the same family with both address and home phone supplied.
- [ ] Row n has home phone, row (n+1) has same family name different home phone - should become a different family.
- [ ] Family member 1 has home phone and address in import. Family member 2 has neither, but matches an existing person. Existing person's family should be updated with new home phone and address.
- [ ] Check matching of existing groups and adding persons to them
- [ ] Check creation of new groups and adding persons to them

tbar0970:
Complete!

PR #442, opened 2018-02-04
Change import to handle duplicates

tim-pearce:
If a person of the same name exists, they are assumed to be the same person and will not be updated except to be added to the group nominated in the import.
Other people being imported as the same family will be added to the existing family.
This and explanation about default group status added to the explanation.
At the end it reports about existing members and families so you can make corrections. (I had one case where a child had been brought along as part of two different families).
This addresses #441.


tbar0970:
Sounds like a good feature.  But this PR includes lots of changes related to other things. Can you have a go at creating a PR just with the changes for this feature?

PR #447, opened 2018-02-07
Database errors on install

### LINKED ISSUE #443: Database errors in initial install
I created a new installation of Jethro using UwAmp and got a series of errors which do not appear to be related to the environment.
I modified the system to report the SQL used to create the attached log. I also made changes as commented.
I will shortly create a pull request for these.
The one I am unsure about is the default congregation print_quantity.

tim-pearce:
Fixes #443


tim-pearce:
Still not sure how this works.
I made changes to make my master the same as yours, but now I see changes I made to match yours listed in the pull request. It's only the last commit that makes it different!

tbar0970:
Since there are only a handful of changes to fix the actual bug here, I'll pull them in manually.

PR #448, opened 2018-02-07
Spreadsheet merge

tim-pearce:
Create a spreadsheet merge function using openTBS.

This allows the creation of spreadsheet templates.
I particularly want to create my own spreadsheets for things like sign in sheets for youth groups.
Also attendance registers.

See call_opentbs_merge_help.class.php for user help and examples.


tim-pearce:
Tom,

Have you had any chance to review this? It is a significant part of the way I use my installation of Jethro, so would prefer it was part of the mainstream rather than having to perpetually having to patch my version.

If you don't have the time, don't like it or whatever I won't say boo hiss or anything like that.

Tim

tbar0970:
Sorry for the delay. Haven't had a chance to look at it properly yet but I think it's a great concept so I expect I'll merge it soon.

Curiosity question:  Are you doing things in these spreadsheets that you couldn't do in a table in a word doc?  Formulas or something like that?

tim-pearce:
Yes, I am using formulas.

Although it is call spreadsheet merge it can handle other document types. The base tool (TBS) is mainly an XML (including HTML) template tool.

I was reluctant to call it document merge because there is already 'mail merge a document'. This tool can do all that does and more, but uses a different syntax.

tbar0970:
Yeah wow, just looking at the openTBS docs, looks very powerful.
Perhaps we should make this the main document-merge tool, and keep the other one as "legacy mode" or something.  I'll have a think.

tbar0970:
Hi Tim. Finally merging this now.  Question for you:  you've added a parameter to person::getCustomMergeData to allow it to return *unformatted* values, and it's this unformatted version that you use in the tbs merge code.  Is there a particular use case that this supports?  For an option field, this will merge the option ID into the results, which I don't think is very useful.

tbar0970:
I suspect it's YYYY-MM-DD dates that you're keen to have "unformatted" in spreadsheet output, which is fair enough. We might need to treat date fields as a special case.

tbar0970:
I've merged this to master, with a few changes particularly around the interface. Still need to resolve the 'unformatted' question above.

tim-pearce:
I'm busy with other things at present. Should be winding down about the 12th.

Issue #344, opened 2017-03-22
SMS sending fails when current user has blank mobile number


tbar0970:
Often an SMS gateway is configured so that the SenderID is set to the current Jethro user's mobile number.  If their mobile number is blank this means a blank SenderID is used, which triggers errors.

Think about how to spot this.

The tricky thing is that sometimes the SMS gateway is configured to send from some fixed SenderID so the lack of mobile number is not a problem.

But requiring someone to have their mobile number entered is not a big deal so we could probably just require it for all SMS sending anyway.

tbar0970:
+1

vanoudt:
Maybe we can tack this onto the permission system... if there's no mobile number associated with a user, assume that they don't have the PERM_SENDSMS permission.

We could maybe add another configuration option - something like "SMS_FROM_NUMBER" that, if configured, is used instead of the current person's mobile number...

vanoudt:
Just submitted a pull request that fixes this: #483

PR #483, opened 2018-05-26
Fix for Issue 344

vanoudt:
Fix for issue #344 

Creates a new system setting, allowing one to use a global mobile number to send SMS messages, rather than the user's number.

Disables SMS sending if there is no global number *AND* the user doesn't have an mobile number.

Also includes a small fix to show users what variables are available in the SMS_POST_TEMPLATE.


tbar0970:
The tricky thing here is that we can't always assume that the SMS_POST_TEMPLATE has been configured to include the _USER_MOBILE_ keyword.

tbar0970:
Thanks for this, but I've gone with a simpler fix in Issue #344

Issue #551, opened 2018-10-19
Note list order is inconsistent if notes created at the same second


GatesheadBen:
If (for example, using an Action Plan) multiple notes are created simultaneously then their order in the notes list is inconsistent.

This is an issue if there is an obvious order to the notes/actions ( 1. Do this first 2. then do this 3. then do this last).

PR #552, opened 2018-10-19
Make note list sorted by created AND subject

GatesheadBen:
Fixes issue #551

### DIFF
--- a/db_objects/family.class.php
+++ b/db_objects/family.class.php
@@ -300,7 +300,7 @@ class family extends db_object
 	function getPostalAddress()
 	{
 		if (!empty($this->values['address_street']) && !empty($this->values['address_suburb']) && !empty($this->values['address_postcode'])) {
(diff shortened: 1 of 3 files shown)


GatesheadBen:
Yes, I thought similarly but then I couldn't decide and/or make work consistently
* Sorting on ID using the uasort as it is an array of arrays
* Sorting as part of the database query and removing the uasort

Your suggestion is certainly more elegant.  I'll have another look when I get to the bottom of the rest of the questions from my end users.

GatesheadBen:
Yes, I'd thought similarly but I couldn't decide/make work consistently:
1) Sorting on ID using uasort as it is an array of arrays (and the ID information is up a level)
2) Sorting in the database query and disabling the uasort

Your idea is certainly more elegant.  I'll have another look when I get to the bottom of the rest of the questions from my end users.  

Their habit is to call the actions "1 Send welcome letter" and "2 ask them to lunch" and I've reminded them that if their list is longer than 9 they will need to pad with a zero.

PR #595, opened 2019-03-19
Generate slides from service runsheets & components

alwr:
This adds functionality to create presentation slides by merging service component or runsheet contents into the XML of an ODP file. ODP files can be opened by powerpoint, google slides, open office, etc.

The template ODP formatting can be customised in powerpoint / open pres to suit and uploaded to documents.
There is 1 new database option for making slides for all service components (prayers, creeds, etc), or only songs.
We've been using this for the last 5 months with great success.  It cuts down our slide build time greatly, the slides are always up to date with the database, and consistently formatted.

Some of the features may be specific to the way I am used to having slides (eg copyright on last slide, slide numbers on every slide, etc).  Happy to incorporate other elements if suggested.  The functions could probably be extended to suit import formats of propitiatory worship software too.


dassaur:
Hi alwr.
Came across your merge scripts - great stuff. My php knowledge is very basic so I tried fixing this issue myself to no avail.
Whenever slides are downloaded using either scripts they are always missing their file extensions, which is very annoying to help a novice understand how it works.
Additionally, (and this is so minor) slides downloaded from the component library only include the filename up to the first space. So if you take for example, 'amazing grace', the output filename would always be 'Amazing'

I wanted to both update you on this issue (hope it's not just me!) and thank you for this!

dassaur:
Upon further testing seems to be a firefox only bug. In chrome it works as expected. In my very limited knowledge I assume it has to do something with firefox interpreting the MIME type.

Anyway, easy fix - tell users to use chrome when building service slides!!

alwr:
Hi @dassaur ,  Glad to hear it might be useful to you!  I built it to make my life easier, but I'm really please if someone else gets some benefit out of it!

Hopefully both issues are fixed with Chrome - I just tested the file name for service components and it worked fine in Chrome 77.0. I do remember something about MIME types being handled differently by some browsers, but don't remember the details.

PR #606, opened 2019-06-11
Communication History

vanoudt:
This isn't intended to be merged yet - more a WIP. The aim is to be able to have a history of communication associated with a person. At the moment, SMS messages are being lumped in with other notes, making it harder to run through actual pastoral notes. In the future, it would be great to have a capacity to see old emails as well. It would also be great to rework the SMS system so that replies can automatically be inserted in as comments.

At the moment, this is just filtering notes according to their subject. It might be better and more performant to make a separate messages class that inherits from notes.

(But, for now at least, this is making notes more useful to me... moving 75 SMS messages out of line of the 2 actual notes on one person makes those notes much more useful!)


vanoudt:
Have now created a new class, based on abstract_note to store communication notes. Also defaulted saving of SMS messages, and shifted them to be stored as as a Person_Comm (Person Communication). 

Next to do is a small script to migrate SMS messages stored in notes to the comms table instead.

Further down the track, would be great to store emails in Person_Comms too.

It would also be good to be able to include replies to sms messages/emails. (This would require implementing a callback, and that might have to be unique to each sms provider. But it would be a really nice feature to have!)

vanoudt:
Wrote the SQL to migrate SMS messages (hardcoded to convert notes with the subject "SMS")

I think this is now at a state where it is merge-able; and more functionality can be added. I'm now running this live, and it's such a pleasure to have communication notes separated out from pastoral notes.

scripts/roster_reminder.php
```diff
@@ -161,7 +161,7 @@
 		require_once JETHRO_ROOT.'/include/sms_sender.class.php';
 		$notification_sms = '';
 		if (!empty($assignees))  {
-			$sendResponse = SMS_Sender::sendMessage($sms_message, $assignees, FALSE);
+			$sendResponse = SMS_Sender::sendMessage($sms_message, $assignees, TRUE);
```
tbar0970: Why do we want the roster reminder to be saved as a person_comm ?

tbar0970:
Hi Nic

I have to apologise, again, for being very slow with pull requests.  I'm looking at them now.

This one looks good, but just has a bit of extra stuff that doesn't belong.  I've added comments inline. Can you adjust and then I'll merge this. THanks

scripts/roster_reminder.php
vanoudt: Hi Tom! Sorry to have taken so long to get back to you on this. I figured that a roster reminder was a communication to a person, and so should be saved to the communication log. Perhaps it would be better to make this optional?

vanoudt: Just threw together an option which would make it optional! :-)

sevasek:

PR #723, opened 2021-09-14
Exclude absentees from rosters

tim-pearce:
Tom,

I don't know what it is but I continue to struggle to understand git and github.

Nevertheless, here is my first set of changes to improve the rostering.
1. Exclude people from the people available if the have been marked as absent for that day in that volunteer group (ie not available) (warning - this might be slow - if it is then I'll revisit how it works)
2. Minor hack to TBS because it was giving an error!
3. Ditto for document_merge when getting the extension. Also slipped in a bit of code to add new lines into the roster (ODS and ODT only)

### DIFF
--- a/calls/call_document_merge_rosters.class.php
+++ b/calls/call_document_merge_rosters.class.php
@@ -11,7 +11,7 @@ class Call_Document_merge_rosters extends Call
 			trigger_error('Template file does not seem to have been uploaded');
 			return;
 		}
-		$extension = @strtolower(end(explode('.', $file_info['name'])));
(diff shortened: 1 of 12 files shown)


tbar0970:
We're on roughly the same track with unavailability.  But I would be cautious about going live with your current code, because it's possible something different will end up in the Jethro core and it will be hard to get back from your branch.

I see a reference to person_group_category for roster views.... what does that do?

tim-pearce:
Alternate implementation already implemented by tbar

PR #748, opened 2022-01-16
'Export check-ins' ignores date parameters

jefft:
If one goes to Attendance -> Checkins, there is a form letting one pick a date range for checkins to export:
image
In Jethro 2.31.1 specifying the date range does nothing. You always get back CSV containing every checkin ever made.

Attached is a small patch fixing the logic in two places: a copy-paste error in the form fields ('from' and 'from' should be 'from' and 'to'), and the SQL should AND the 'venue' clause with the date clause.


tbar0970:
Thanks for the diagnosis and fix :)

PR #816, opened 2022-08-24
Resilient db object values

vanoudt:
Had a real world issue here where a person had no history associated with them - not at all sure how that happened, but it did! This caused all sorts of headaches! Trying to edit the person resulted in an Out of Memory error!

Unserialising the history was returning an empty string, rather than an empty array. This fixes that.

It might be preferable to put a note in there that something has gone wrong - although as soon as new history is added, it no longer causes any issues

### DIFF
--- a/include/db_object.class.php
+++ b/include/db_object.class.php
@@ -3,1181 +3,1180 @@
 class db_object
 {
 
-	public $id = NULL;
-	public $fields = Array();
-	public $values = Array();
+    public $id = NULL;
+    public $fields = array();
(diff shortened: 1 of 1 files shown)


tbar0970:
Thanks for this. I decided to fix it a bit differently in #825

tbar0970:
Sorry, I claimed on 30th Sept that I'd committed some other fix for this, but I can't for the life of me see what I actually did. Maybe I imagined it.  I will merge your change now.

PR #375, opened 2017-06-24
Attendance auto headcount

vanoudt:
Hi Tom,

A quick pull request to scratch an itch of mine. This simply automatically increments or decrements the headcount for the current congregation/group when doing attendances. The value can still, of course, be changed manually. 
If the value is changing from Present->Unknown or Present->Absent, it decrements it by one.
If the value is changing from Unkown->Present or Absent->Present, it increments it by one.

When we have guests, I want to include them in the headcount. Up till now, I've been doing the attendance, hitting save to see the total members present, and then going back into attendance and adding in the number of visitors present. Doing it this way, I can mark everyone present or absent, and then manually add in the number of guests that I've counted...

Issue #704, opened 2021-03-22
Attendance running total


kristanslack:
It would be really help to have a running total when attendance is being marked. That way, when we know that there were N visitors, we could simply add that to the running total - rather than needing to save and then go back in.

That is, we don't do a headcount (kinda relying on Jethro to do that after marking people present) - but we are always aware of how many visitors/guests are there.

s4069b:
I really like this feature in v2.34.0. Thank you!

s4069b:
Actually, after playing around a bit more, I just noticed a minor formatting glitch. If the Congregation/Group you're marking attendance for does not have long names, then this word-wrap happens. 
Not sure how to stop the text wrapping. Might have a look later. 

<img width="670" alt="Screenshot 2023-08-29 at 8 46 06 am" src="https://github.com/tbar0970/jethro-pmm/assets/29622389/2856851b-8287-4119-83c0-3b380189a119">

s4069b:
This 'fixes' the formatting. Hope nothing else broke?

db_objects/attendance_record_set.class.php
change line 357 from class="muted" to class="muted nowrap"

![Screenshot 2023-08-29 at 10 17 51 am](https://github.com/tbar0970/jethro-pmm/assets/29622389/991c37a1-e5d1-400a-a7a8-367f9b8df172)

PR #837, opened 2022-12-08
Auto-increment headcount when recording attendance

brett-mcgirr:
Hi,

I've got a PR similar to #375 but much smaller so should hopefully be easier to verify.

It's purpose is to add up those present/absent on the page so that the number is immediately available, and it can also be used to add guests once the regulars are checked off.

Cheers,

Brett

### DIFF
--- a/db_objects/attendance_record_set.class.php
+++ b/db_objects/attendance_record_set.class.php
@@ -343,7 +343,7 @@ class Attendance_Record_Set
 			$headcountValue = Headcount::fetch('person_group', $this->date, $this->groupid);
 		}
 		?>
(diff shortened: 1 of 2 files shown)


vanoudt:
Brilliant! I've been manually pulling in/updating #375 for about 5 years now on our system, and think it doesn't apply/work properly these days. Would love to have this functionality added! Simpler is normally better! :-D

tbar0970:
I'm sorry for neglecting this PR for so long.
In https://github.com/tbar0970/jethro-pmm/issues/704 I've now added a display of the number-marked-present next to the headcount box, which I believe satisfies the same need.
So I'm going to close this PR. Let me know if there's some need that issue #704 doesn't meet.
image

brett-mcgirr:
Brilliant, thank you.

PR #840, opened 2022-12-10
Load person history column

### LINKED ISSUE #836: Editing family details gives fatal error: Allowed memory size of  bytes exhausted
Hi,

When updating a family name, we're getting an error saying allowed memory size exhausted. The code in question is the following:

include\db_object.class.php
Line: 358

356 $user = $GLOBALS['user_system']->getCurrentPerson();
357 $now = time();
358 $this->values['history'][$now] = 'Updated by '.$user['first_name'].' '.$user['last_name'].' (#'.$user['id'].")\n".implode("\n", $changes);

$now is set to some unix time, e.g. 1670462680 and when it's used as an index into the string $this->values['history'], PHP is trying to allocate a string that is 1,670,462,680 characters long which is too much. 

I'm not 100% sure what the intention of the $now value is, but might the following code be the way forward?

358 $this->values['history'] = 'Updated by '.$user['first_name'].' '.$user['last_name'].' (#'.$user['id'].")\n".implode("\n", $changes);

Cheers.

brett-mcgirr:
Fixes #836 

Explicitly load the person history column so that the history value is an array and not an empty string.

### DIFF
--- a/db_objects/person.class.php
+++ b/db_objects/person.class.php
@@ -831,7 +831,7 @@ class Person extends DB_Object
 	function getInstancesQueryComps($params, $logic, $order)
 	{
 		$res = parent::getInstancesQueryComps($params, $logic, $order);
-		$res['select'][] = 'f.family_name, f.address_street, f.address_suburb, f.address_state, f.address_postcode, f.home_tel, c.name as congregation, ab.label as age_bracket';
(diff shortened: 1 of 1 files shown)


tbar0970:
Thanks for this. I decided to fix it a different way, to avoid the performance impact of fetching the history column in scenarios where it's not needed

PR #851, opened 2023-01-23
Set X-Jethro-User header, etc, to identify user operations in webserver logs

jefft:
In webserver access logs, it is useful to see which Jethro user made each request. This patch makes that possible to adding 3 new HTTP response headers:

- `X-Jethro-User` is the username (for staff) or email (for members)
- `X-Jethro-UID` is the internal ID. This is consistent across staff/member logins which might make it preferable to `X-Jethro-User` (also avoids logging PII).
- `X-Jethro-Session` is a subset (for privacy) of the session id

These could be used in an Apache `LogFormat` directive as follows:

```
LogFormat "%v:%p %a %{X-Jethro-Session}o %{X-Jethro-User}o %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\" %D uid=%{X-Jethro-UID}o" vhost_combined_timed
```

Note: it is also possible to pass 'notes' from PHP to Apache:
```
apache_note('jethrouser', array_get($_SESSION['user'], 'username'));
```
then logged with `%{jethrouser}n`. However `apache_note()` doesn't work with php-fpm, or other webservers like nginx, so headers are better.

### DIFF
--- a/include/system_controller.class.php
+++ b/include/system_controller.class.php
@@ -81,6 +81,8 @@ class System_Controller
 
 	public function run()
 	{
+
+		$this->setGlobalHeaders();
 		if (!empty($_REQUEST['call'])) {
 			$this->initErrorHandler();
 			$call_name = str_replace('/', '', $_REQUEST['call']);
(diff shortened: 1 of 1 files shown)


tbar0970:
I can see the value of having these details in the webserver log.  But it seems a pity to add more weight to every http response.  Is there no other way to be able to pass stuff to the log? (I note the limitations of apache_note)

Also, it looks like your code for the UUID header is actually using the *username* for control centre accounts rather than person ID

tbar0970:
If headers are the only way to achieve the logging we want, I'd argue for just one header with personid.

jefft:
Headers seem to be the only way (see e.g. [here](https://stackoverflow.com/questions/13396740/need-apache-note-function-with-nginx-usertrack-for-nginx). In Atlassian apps like Jira, these headers are named `X-AUSERNAME` and `X-ASESSIONID`,

These headers aren't sent for `/resource/*` queries, just *.php, so bandwidth is trivial. If anyone cared they could prevent these headers being sent on to the user with e.g.:
```
Header always unset "X-Jethro-User"
```

Why 3 headers: `X-Jethro-UID` is consistent across member/staff logins, and doesn't leak PII in logs, but I'd guess 90% of users won't care about either of those things, and will prefer `X-Jethro-User` for readability and log greppability. There's no harm in having both in logs.

`X-Jethro-Session` allows debugging of painful situations where a user is logged in on multiple devices or browsers. It allows one to identify a stream of requests coming from just one device.

PR #791, opened 2022-05-19
Document merge updates

tim-pearce:
Implements most of #743 plus a few issues with php 8.1 (#790)
Still inefficient in how it handles attendance data.

Issue #790, opened 2022-05-19
php 8.1


tim-pearce:
Working on improvements to document merge I have been bitten by deprecations in php 8.1 particularly builtin function that will no longer accept nulls in place of a string.
In particular strpos, str_replace and htmlspecialentities.
This last one had me baffled as it pointed to line 122 of general.php even when the code was moved. Eventually I realised that the exception report, complete with all the formatting had been written to service notes in the database.
Some of these issues will be resolved when I submit my document merge changes/

tim-pearce:
I've done a run through my dev system and so far have only come across 
Line 16,19 & 37 of File .../include/paginator.class.php
  Implicit conversion from float 65.44444444444444 to int loses precision
Changing 
$this->_slice_size - 1 
  to 
round($this->_slice_size - 1)
fixes the issue.
Using round() is consistent with the existing code on line 40.

jefft:
Yes, Ubuntu 22.04 defaults to PHP 8.1, the Jethro installer page looks sad:

![image](https://user-images.githubusercontent.com/205995/189380656-6397292c-5b5a-4032-865a-461e1de1f0c2.png)

tbar0970:
Fixed in #2cc6f45

Issue #743, opened 2021-12-14
Document merge enhancements


tim-pearce:
I have just updated my test environment to version 2.31.1 after applying changes I have made in my system.

Most of my 'hacks' are related to document merge. In due course I will construct a pull request when I apply these in my development environment. I will list my changes here. Some of them are not in a suitable state for the 'real world' yet.
call_document_merge.class.php
- If the passed template is a .html I display the data using print_r - this enables me to see what template variables are available - not very user friendly in this form!
- For a person query, include 'extended' items (eg a list of groups) which are generated for the query. Currently only works if 'Person ID' is one of the selected fields for the report.
- Replace newlines with spaces in strings
call_document_merge_rosters.class.php
- Provide a 'role_cr' field which for some document type (ODS and ODT) inserts the correct code to make each item go to a new line (carriage return and/or new line does not work inside a document)
bulk_actions.template.php
- Pass the queryid to document_merge
view_6_attendance__2_display.class.php
- The attendance figures are passed to document_merge_rosters in a different, more efficient way

In addition I believe I should turn document_merge around so it gets it's primary data and sequence from the query and then tacks person/family data to that. I currently use a hack of the printCSV process that requires the query to include 'person ID'. I need to modify printCSV so it can return an array which is keyed on person/family ID

tim-pearce:
I have made significant progress on this.
Incorporates the latest version of TBS which is compatible with PHP 8.1

tbar0970:
A pull request would be great :)

tbar0970:
Re

> In addition I believe I should turn document_merge around so it gets it's primary data and sequence from the query and then tacks person/family data to that. I currently use a hack of the printCSV process that requires the query to include 'person ID'. I need to modify printCSV so it can return an array which is keyed on person/family ID

Document merge isn't always triggered from a report/query - bulk actions can also be launched from various other places where persons are listed.  Also, the user may tick some persons but not others in the list.

tim-pearce:
Not ready for the pull request just yet - hopefully next week or maybe the one after.

I think I may have already worked out that it needs to be driven by the tick list and that's how the revised code continues to work.

I'm using PersonQuery (don't have the code in front of me) to get the query data. The options in the GetResults function are now 'html', 'csv' and 'array'. 'array' returns the data as an array - based closely on the 'csv' function.

[...]

PR #857, opened 2023-02-07
Doc merge updates

tim-pearce:
This pull request supersedes #791

Implements most of #743 plus a few issues with PHP 8.2 (#790) Still inefficient in how it handles attendance data.

- incorporates a recent version of TBS which works for PHP 5, 7 & 8 (I had to make changes for it to work with PHP 8.2)
- includes a 'Dump' facility which shows ALL fields available
- allows for storing templates on the server when used for rosters and attendance
- brings in 'extended' fields in reports
- provides separate tables for each member status (both congregation and group) where they exist
- help and examples are updated and extended
- rosters now includes a list of people on the roster

Tested on PHP 8.2 and 5.6


tim-pearce:
Tom,
Any chance you could move this along? It's quite a bit of code for me to keep merging into Jethro whenever there is an update.

tbar0970:
Hi Tim. I understand how applying your patches is a pain, and I'm sorry I've left these PRs unresolved for so long.  The challenge for me is that as project maintainer I need to review code before I 'take responsibility' for it in the main branch.

This PR has 23 changed files (8 of them are binary files - ODT examples etc). That's a lot of files to review.   It's also harder when there are several problems/enhancements being addressed in the one PR.  
For example I'm not clear how call_csv relates to document merge.

To give me a better chance of being able to understand and merge these, could I get you to submit some separate PRs for specific things.  For example your changes to general.php and the paginator class look like standalone issues - if you can submit a PR for each of those, with just one file changed,  and an explanation of the problem it's solving, I can merge those quickly.  And then I'll have a better chance of getting my head around a document-merge PR which has more extensive code but no other distractions.

Sorry to be a pain, but realistically this is the only pathway that I can see working, given the limited time I've got to review these.

tim-pearce:
Tom, thanks for the feedback. I will break it down and add more explanations.
Some of those seemingly unrelated changes are because I am on PHP 8.2

tim-pearce:
Superseded by latter requests

PR #864, opened 2023-02-08
Document merge for reports and attendance

tim-pearce:
call_document_merge.class.php
Add 'dump process
Create tables by group and status
Use person_query to get the data
bulk_actions.template.php
Add the 'Dump' option to the dialog
Pass the queryid if available
views/view_6_attendance__2_display.class.php
Pass attendance data in a different way and include group status info

### DIFF
--- a/calls/call_document_merge.class.php
+++ b/calls/call_document_merge.class.php
@@ -15,8 +15,9 @@ class Call_Document_Merge extends Call
 
 		$file_info = array_get($_FILES, 'source_document');
 		$content = null;
-		$template_filename = null;
-		if (!empty($file_info['tmp_name'])) {
+		if (array_get($_REQUEST, 'template_format') == 'dump') {
+			$extension = '@@@debug@@@';
+		} else if (!empty($file_info['tmp_name'])) {
 			if (!empty($_REQUEST['save_template'])) {
 				$ok = TRUE;
 				if (!is_dir(self::getSavedTemplatesDir())) {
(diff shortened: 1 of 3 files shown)


tbar0970:
Can you explain more about the difference between merging report results, vs merging the persons in the report?  How is this different to just exporting the report itself as csv or whatever?

tim-pearce:
By including all the person data alongside the report results the template has all the information about the person not just what is in the report.
Previously I was only presenting the person data and missing out on other items provided by the report!

tbar0970:
Hmm OK.  So previously, when you viewed a report, selected all the persons, and chose "merge a document" from the bulk actions at the bottom, your template could only include the standard person fields (first_name, last_name etc).  This adds the ability to also include columns in the report (eg "which groups the person is in", "adult family member names" etc) as well.  Have I got that right?

tim-pearce:
Replied 9/2/23
Yes.

tbar0970:
Hi Tim
Are you currently using the keywords that are prefixed with person status?
image
I find this approach quite confusing - the prefix in the keyword actually functions as a filter.  It also blows out the number of possible keywords hugely.
I'd prefer to think about alternative ways to add filtering capability if it's needed.

tim-pearce:
I definitely use this.
It's more of a 'group by' function than a filter.
In the case of our attendance recording we have 'regulars' and people who we have started marking but haven't yet become regular attenders so we want the reported separately on the same worksheet.
For a list of members I would want to report each group of member categories separately but on the same report.
I will raise a suggestion to expand this functionality to allow the user to specify a field to group by.

tbar0970:
OK. So if you were just viewing it as a normal report, you'd be using the "group the results by..." option to divide them up into regulars vs newcomers etc - correct?
Now that the document merge is aware of the report, it could use that same grouping field ?

tim-pearce:
That would be true for reports but not for attendance. Am creating a new issue in the other browser tab now.

tbar0970:
Question:  Are you just using the status-prefix when merging *attendance* ?
In that case we could disable it when merging from a report, to avoid the bloat.

PR #865, opened 2023-02-08
Document merge for rosters

tim-pearce:
all_document_merge_rosters.class.php
Create temporary document name
Allow for 'Dump' option
roster_view.class.php
Pass more data, especially in relation to the services views/view_7_rosters__1_display_roster_assignments.class.php Add in the dialog like the other two

### DIFF
--- a/db_objects/roster_view.class.php
+++ b/db_objects/roster_view.class.php
@@ -422,8 +422,8 @@ class roster_view extends db_object
 
 		$to_print = Array();
 		foreach ($services as $id => $service_details) {
-			$to_print[$service_details['date']]['service'][$service_details['congregationid']] = $service_details;
-			$to_print[$service_details['date']]['service'][$service_details['congregationid']]['id'] = $id;
+			$to_print[$service_details['date']]['service'] = $service_details;
+			$to_print[$service_details['date']]['service']['id'] = $id;
 			$to_print[$service_details['date']]['assignments'] = Array();
 		}
 		foreach ($this->getAssignments($start_date, $end_date) as $date => $date_assignments) {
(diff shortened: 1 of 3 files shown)


tbar0970:
I'm confused about what this actually does.
Can you post an example of what you produce using this process?
It's completely distinct from the generic person-document-merge process, right?

tbar0970:
(I've seen the sign-in-sheet template, but I'm confused as to how it relates to a multi-column roster)

tim-pearce:
I replied from the email with attachments? Got to go to a meeting now.

tbar0970:
Email attachments might not make it through to github. Might need to try uploading direct in the web UI.

tim-pearce:
For youth ministries, sign-in/out is a requirement for 'child safe' so I produce a sheet with the names of everyone rostered on that day to sign.
The other thing is a printed roster to put on the notice board or for people to take.
(All names removed except mine for privacy).
[roster.ods](https://github.com/tbar0970/jethro-pmm/files/10703166/roster.ods)
[Sign_in_out_sheet.ods](https://github.com/tbar0970/jethro-pmm/files/10703168/Sign_in_out_sheet.ods)

tim-pearce:
My comments are not coming through email. Maybe it's because I send using a different email address! 
Replied 9/2/23
It's all the people rostered on on that date (or dates).

db_objects/roster_view.class.php
```diff
@@ -422,8 +422,8 @@ public function printCSV($start_date=NULL, $end_date=NULL, $return=FALSE)
 
 		$to_print = Array();
 		foreach ($services as $id => $service_details) {
-			$to_print[$service_details['date']]['service'][$service_details['congregationid']] = $service_details;
```
tbar0970, line 425: There's a problem here.  A roster view may include details of several services on the same date (for different congregations). The array needs to be indexed by congregationid or they will overwrite each other.

db_objects/roster_view.class.php
```diff
@@ -460,6 +460,34 @@ public function printCSV($start_date=NULL, $end_date=NULL, $return=FALSE)
 
 		foreach ($to_print as $date => $ddetail) {
```
tbar0970, line 461: likewise here, it's assuming that there is one service per date, but that's not the case

tim-pearce:
I haven't really been using congregations except to separate out rosters for week day ministries, so not surprising I missed that.
Will revisit that code and keep that in mind for other code.

db_objects/roster_view.class.php
tim-pearce: Trying to wrap my head around this.
I've experimented with this on my system and it works with two congregations with services on the same days.
I think it works because getCongregations looks at the members and all the members in the service belong to the same congregation..

tim-pearce:
The purpose of getting the service details is as a nice to have feature to allow service details to be shown on the roster.
I think maybe it should stay as is.

db_objects/roster_view.class.php
tbar0970: Ah yes, I think I misread  this line - sorry.  It's already indexed by congregationid.

tim-pearce: I've changed the order of the indexing and made service details available from all included congregations now.
The code in this PR was pretty stale so I have created a replacement PR #932 after rebasing my fork against your latest version.

PR #924, opened 2023-06-24
Update service.class.php to include tooltips (with ccli# and comments) for songs (as service components) in a roster view

s4069b:
I have not submitted a pull request for a very long time, so I hope this one makes sense. 

Be aware that the ccli number and comments are retrieved using the song title - which may not be unique. 

And I also changed nbsp($res) to $res - otherwise html code for the tooltip got 'lost'.

### DIFF
--- a/db_objects/person_group.class.php
+++ b/db_objects/person_group.class.php
@@ -311,6 +311,13 @@ class Person_Group extends db_object
 			return FALSE;
 		}
 
+		$roles = $GLOBALS['system']->getDBObjectData('roster_role', Array('volunteer_group' => $this->id));
+		if ($roles) {
+			$role = reset($roles);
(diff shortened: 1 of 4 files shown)


tbar0970:
You're on the right track.  Few adjustments needed:

You're right that looking up the details by title isn't a good idea.
I am pretty sure that $item actually includes the data you want already - no need to run queries.
Add bam($item) to have a look and check.
Also, using the title for the element ID isn't such a good idea. 
id="item<?php echo (int)$item['id']; ?>"
would be better

The nbsp function is needed to make sure the titles don't get squished when there's a wide table.  You could add it back just on $title.

Also: If there's no CCLI number or comments, it'd be better not to show a link

Also: comments might be multi-line, so you might need to work on the layout to cope with that.

Issue #918, opened 2023-06-20
Choose to display CCLI number and Key with Song Name in Roster View (service component)


s4069b:
We have setup a 'roster view' which includes 'songs' (service component). This is used to generate a roster reminder to the musicians. We find that the CCLI number is helpful in communicating which song we're after (ie not just the song title). And a common question the musicians will ask is - 'what key is that in?' 

There is no option atm to include the CCLI number or Key in a Roster View when adding Songs as a service component.

Is this a feature others would use? I can have a mess around.

At the moment our work around is to add the CCLI number in brackets as part of the Song name. We also add the key in which we anticipate it will be played. But this workaround means we have two CCLI numbers and two keys for each song component - which will no doubt cause confusion down the track.

tbar0970:
The "comments" field was designed to be the place for musical details like key, tempo, links to recordings and charts etc.  

Would it be sufficient to have the comments and CCLI number as a click-to-show ?  Or do you need it to show up in a printed roster etc?

s4069b:
Thanks for the quick reply. I can see the "comments" field in the service component (ie if I pick a song, and edit it) - this would allow me to add the ccli and key in the comments for that song. However there's a double-up then, because there is already a spot to enter the CCLI number. 
Also, I can not work out how to have the comments for a song appear in the "roster view" where I have included songs as a 'service component' in the roster.

tim-pearce:
I am hoping to add a document merge option to the service sheet view. This would give the ability to include whatever is available on your personalised version.

tbar0970:
This issue is about how it shows up on a roster, not the service sheet.

tim-pearce:
There would be multiple songs in a service, would there? So are we after a list of some song details to be available in the corresponding roster view? 
In the document merge for rosters that I've created I pull some details about the corresponding service in -
[roster.date] = 25 Jun 2023
[roster.format] = 
[roster.topic] = Sunday service
[roster.notes] = 
[roster.notes_cr] = 
[roster.comment] =

s4069b:
Just after the ability to add CCLI number and Key alongside the Title of each song in a roster view - where the songs are added to the roster view as a service component. I have not had time to have a look at it yet.

s4069b:
OK. I tried having a look at this but I'm very rusty. If someone could point me to the right area, that would be handy. 
I'm thinking perhaps the way forward/goal would be to add another option in the dropdown here:
".. All Songs including CCLI # and Comments" 
[...]

PR #926, opened 2023-06-27
Update service.class.php to include ccli&comments tooltip

s4069b:
Here's a second go at satisfying issue/feature request "Choose to display CCLI number and Key with Song Name in Roster View (service component) #918"

This time the Ccli number and comments are retrieved using 'componentid' (discovered using bam($this)) .. so it is unique this time.
The comment length is limited to 20 characters (could choose longer easily enough)
The tooltip does not appear if there is no ccli number or comment associated with the song service component
nbsp() is applied to the title (for formatting). 
and componentid (ie an integer) is used for the tooltip id (instead of 'title)

Have I missed anything?

### DIFF
--- a/db_objects/service.class.php
+++ b/db_objects/service.class.php
@@ -377,10 +377,34 @@ class service extends db_object
 					$compCatID = (int)substr($fieldname, 6);
 					$res = Array();
 					foreach ($this->getItems(FALSE, $compCatID) as $item) {
-						$res[] = ents($item['title']);
+				        $title = nbsp(ents($item['title']));
+ 				        $songid = ents($item['componentid']);
+                			$sql = 'SELECT ccli_number FROM service_component WHERE id = "'.$songid.'"';
+		                	$ccli_number = $GLOBALS['db']->queryOne($sql);
+ 		                	$sql = 'SELECT comments FROM service_component WHERE id = "'.$songid.'"';
+                       			$comments = $GLOBALS['db']->queryOne($sql);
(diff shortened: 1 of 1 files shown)


s4069b:
also, 'show print version' includes the ccli and comments ... which I actually prefer.

tbar0970:
Thanks.
I did a bit of further work on this.
I'm keen to keep the "printable" version clean of details.
But they still show up in the public area and members area.

s4069b:
Just had a look. Makes my attempt look like the hack it was😀 and I learned lots looking at your code. 
Thanks for taking time to fix/redo.

tbar0970:
Thanks for getting this going, it will actually be useful for my music team
too.

s4069b:
I don’t play music so this may not be helpful. How often does the desired key change? And are there other details in ‘comments’ which musicians want access to?  If so, then perhaps an additional link in the ‘tooltip’ to edit the song might be added? It could be a time saver (rather than needing to exit the roster and come back in via service component menu)

tbar0970:
Yeah I was just thinking about this. 
I wouldn't say keys change often, but there are other details like a link to the relevant youtube version of a song, links to charts etc, which can be developed over time.
A link to edit could be good. We just wouldn't want it to take up too much space.  Maybe we could fit an icon-wrench in there.

tbar0970:
I've added a little wrench icon to link to edit the component.

s4069b:
Nice!

PR #932, opened 2023-08-08
Document merge for rosters

tim-pearce:
Replaces PR #865
Now copes with multiple congregations
I had to strip out the SizeDetector test on my system as it caused the option to be omitted.


sevasek:

Issue #947, opened 2023-09-05
Temporary files not deleted


jefft:
Jethro is not great at cleaning up its temporary files. I'm seeing:

```
contactlistXXXXXX
contactlistphotoXXXXX
contactlistphotoXXXXX.jpg
jethro-mc-XXXXXXXXXX
mailchimp_contentXXXXXX
phpXXXXXX.docx
phpXXXXXX.odt
PhpWordXXXXXX
```

Where X is a random letters/number.

E.g. every time I download a .docx of a family contact list (Families -> Contact List -> <show results> -> Download As: DOCX), a new /tmp/contactlistXXXXXX temp file gets created.

tbar0970:
Do you think your PR cleans up most of these, so I can close this issue?

jefft:
The PR doesn't fix most of the tempfiles, but it does fix the `contactlistXXXXXX` ones, which are the largest (as in #949). I'd say the severity is low.

jefft:
Added a PR to delete `/tmp/jethro-mc-TSTAMP` directories that accumulate whenever Mailchimp starts failing.

PR #948, opened 2023-09-05
#947 - avoid double-creating temp files

jefft:
The PHP `tempnam` function creates an empty file, which isn't cleaned up by code like this:
```
$tempfile = tempnam(sys_get_temp_dir(), 'contactlist').'.docx';
....
unlink($tempfile);
```

### DIFF
--- a/views/view_0_send_mc_campaign.class.php
+++ b/views/view_0_send_mc_campaign.class.php
@@ -61,7 +61,7 @@ class View__Send_MC_Campaign extends View
 			if ($attachment_error) return; // Do not send.
 
 			$zip = new ZipArchive();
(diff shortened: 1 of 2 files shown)


tbar0970:
Hmm, I see you've stopped Jethro from sticking an extension on the end. But even like this, I don't believe PHP will clean them up itself:

https://stackoverflow.com/questions/72057808/do-i-need-to-cleanup-files-created-by-tempnam-myself

jefft:
There's already code later on to delete `$tempfile` (or whatever the var is called), which I left untouched.  The problem is that the code:
```
$tempfile = tempnam(sys_get_temp_dir(), 'contactlist').'.docx';
```
produces two tempfiles, `/tmp/contactlistXXXXXX` and `/tmp/contactlistXXXXXX.docx`, and only the latter (`$tempfile`) is cleaned up.

tbar0970:
Aaaah cool

PR #954, opened 2023-09-07
Remove underlines from action links

jefft:
I think this UI tweak, removing underline links, makes the UI a bit cleaner. WDYT?

Before/after screenshots:

#### Before
image
#### After
image

#### Before
image
#### After
image

#### Before
image
#### After
image

### DIFF
--- a/resources/less/jethro.less.php
+++ b/resources/less/jethro.less.php
@@ -567,7 +567,6 @@ ul {
 	opacity: 0.4;
 }
 #body a:not(.label), .modal a, .clickable, button.btn-link, input.btn-link {
-	text-decoration: underline;
 	color: @linkColor;
 }
 #body a.btn, .modal a.btn {


tbar0970:
Hmm, I get the thinking, but I'm not sure about removing every underline from every link.  I'll have a play with it.

tbar0970:
I do like removing underlines from the links within headings, as in your first screenshot.

And I can see how when there is an icon (eg spanner), it's obvious that the link is a clickable action, so underline is not needed for that.

I notice there are now some options we could play with to make the underlines less pronounced:
text-decoration-color: #bbb 
or text-decoration-style: dotted

Another option, on desktop, is to only show the underline on hover.

I've just realised Jethro is a bit inconsistent with lists of objects.  In the group list shown in your screenshots, all the group names are (underlined) links to view the group.  But elsewhere (eg Persons > list all) we have non-linked names, with a "view" link on the side, which makes for less underlining overall. There is some JS magic so that in such a situation you can still click anywhere in the table row to view the object (but this is missing for the notes list!).  We should probably be more consistent with this.

Issue #913, opened 2023-06-14
Refactor document merge classes


tim-pearce:
Now that I've created a few of these and I have one in the wings for contact list and thinking about one for service run sheets I begin to see patterns that could be better put into a base class.
I propose to create a base class and change the various document merge classes to use it.
This would probably also include the dialog.
My question is - where do I put the base class? In calls or includes or somewhere else?

tbar0970:
If your abstract class extends call, put it in the calls folder and call it abstract_call_xyz.class.php

tim-pearce:
There aren't any other files named like that in that folder. There is one such file in the includes folder?!

tbar0970:
There are some abstract db-object classes in the db-objects folder, and
some abstract view classes in the views folder, so an abstract call class
should go in the calls folder.

tim-pearce:
I'm ok with that.
I'll probably call the class abstract_Document_Merge because I've already used the name Document_Merge.

PR #966, opened 2023-10-08
Create abstract_call_document_merge.class.php

tim-pearce:
Issue #913
This is a new abstract (or base) class for document merge. Will make no change to current document merge classes but will be required for subsequent updates. Incorporates an update to the 'preview' function added to the most recent document_merge class.


sevasek:

PR #967, opened 2023-10-08
Update call_document_merge.class.php

tim-pearce:
Update to call_document_merge that relies on the abstract class.


sevasek:

PR #968, opened 2023-10-08
call_document_merge_rosters

tim-pearce:
This version requires the abstract class
Resolves the issue of two congregations with services on the same day Replaces PR #932
As previously I had to strip out the SizeDetector test on my system as it caused the option to be omitted!


sevasek:

PR #1044, opened 2024-06-17
Fix #1043 - attendance percentage filtering excludes 0% attendees

### LINKED ISSUE #1043: Person Report 'attendance percentage' filter omits 0% attendance people altogether
Jethro's Person Reports has a the ability to find people whose attendance is less than X percent over the last Y weeks:

image

Say we filter for under 50% attendance over the last 2 weeks, as in the screenshot.

Jethro has a bug where if someone hasn't attended at all in the last 2 weeks, they are filtered out out of the report. Their attendance is 0%, which is certainly less than 50%, so they should be displayed, but they're not.

To replicate, find someone who hasn't attended in the last 2 weeks, note their congregation, and create a person report as above. You'll see that person isn't shown in the results.

jefft:
altogether


jefft:
As Tom explained offline, Jethro's current behaviour is correct. If there are no attendance records in the last 2 weeks, we cannot assert that someone attended 0% - we just don't know. So we filter out that user.

PR #1049, opened 2024-06-21
Only display people stats if the roster view has some Roster Roles (vs. Service Fields). Fixes #1048

### LINKED ISSUE #1048: Roster 'people assigned' analysis breaks if no Roster Roles are present
Under Rosters -> Display Roster Assignments, Jethro now displays some stats about assignees (unless `ROSTERS_HIDE_ANALYSIS` is is defined):

image

This section breaks if a Roster View contains only Service Fields, and no Roster Roles:

image

The logged error is

```sql
SQLSTATE[42000]: Syntax error or access violation: 1064 You have an error in your SQL syntax; check the manual that corresponds to your MariaDB server version for the right syntax to use near ')
                                ) x
                                GROUP BY personid, first_name, last_name
                                HAVING date_count ...' at line 9
Line 171 of /home/jethro/code/2.35.1/app/include/jethrodb.php
```

To replicate, go to  Rosters -> Define Roster Views, and define a minimal view containing only Service Fields:

image

Then view your new roster.

jefft:
I've also moved the <h4> header into the function so that it can be omitted if `$this->getRoleIDs()` is false. It looks like it should be in there anyway.

### DIFF
--- a/db_objects/roster_view.class.php
+++ b/db_objects/roster_view.class.php
@@ -753,10 +753,12 @@ class roster_view extends db_object
 						<?php
 						echo '<strong>'.str_replace(' ', '&nbsp;', date('j M y', strtotime($date))).'</strong>';
 						if (!$editing && !$public) {
+							$names = Array();
 							$emails = Array();
 							$mobiles = Array();
 							$personids = Array();
 							foreach ($ddetail['assignments'] as $roleid => $assignees) {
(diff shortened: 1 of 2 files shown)


tbar0970:
Sorry I didn't merge this before doing other work which now conflicts.
I'll apply a fix myself.
BTW there's some CEC-specific hacks in this PR.

jefft:
Thanks. Sorry I mucked up the PR - git and I are still not friends.

Issue #1035, opened 2024-06-03
Altering person status options


tbar0970:
In system config page you can edit person status options.
They are shown as just a bunch of textboxes, no ID numbers or anything.
I wanted to re-order the options so I copied and pasted various values.
But this had the effect of changing people's status, because it's indexed by number.

So in the short term we should add numbers so it's more clear that renaming options will affect people who already have that status selected.
And as a FR we should make it possible to re-order existing statuses.

Really the right solution is to introduce a person_status table.  That's plenty of work though.

jefft:
> Really the right solution is to introduce a person_status table. That's plenty of work though.

Yes, a `person_status` could basically just copy `person_group_membership_status`:
```sql
MariaDB [je_demo]> desc person_group_membership_status;
+------------+---------------------+------+-----+---------+----------------+
| Field      | Type                | Null | Key | Default | Extra          |
+------------+---------------------+------+-----+---------+----------------+
| id         | int(11)             | NO   | PRI | NULL    | auto_increment |
| label      | varchar(255)        | NO   | UNI | NULL    |                |
| rank       | int(11)             | NO   |     | 0       |                |
| is_default | tinyint(1) unsigned | YES  |     | 0       |                |
+------------+---------------------+------+-----+---------+----------------+
4 rows in set (0.001 sec)
```
Then we could have referential integrity for `_person.status`.

tbar0970:
Made good progress on the full solution.
Todo: interface for editing the options.
Todo: Make sure they are ordered by rank in report results and report groups.

tbar0970:
Todo: foreign key in the upgrade script and init sql.

tbar0970:
todo: search the code for references to 'archived' and 'contact'.

tbar0970:
- [x] Todo: foreign key in the upgrade script and init sql.
- [x] Todo: interface for editing the options.
- [x] Todo: foreign key in the upgrade script and init sql.
- [x] todo: search the code for references to 'archived' 
- [x] search the code for 'contact'.
- [ ] Make sure statuses are ordered by rank (not name) when person reports are ordered or grouped by status.

tbar0970:
This is complete, but it's quite a wide-ranging change, so we should be on the lookout for side effects.

jefft:
See PR #1083 which fixes reports saved with the old statuses.

tbar0970:
TODO: auto-upgrade. NOte to self.

PR #1083, opened 2024-09-28
Fix person_reports and action plans after a Jethro 2.36+ upgrade

jefft:
Jethro's `person_status` table contains serialized strings representing parameters for reports. For instance, `person_status.params` might start with:

```
a:17:{s:5:"rules";a:3:{s:8:"p.status";a:1:{i:0;s:1:"0";}
```
The representation of `p.status` has changed in Jethro 2.36 (see #1035). No longer is it a string (`"0"` in the serialization above), but an integer foreign key into `person_status`. The correct equivalent in this example would be:
```
a:17:{s:5:"rules";a:3:{s:8:"p.status";a:1:{i:0;i:1;}
```
This change cannot be done in SQL. We need some PHP code to deserialize each `p.status`, fix the values and re-serialize it. 

This script is based on `scripts/fix_reports.php`, and placed in `upgrades/` following the naming scheme there, so hopefully upgraders see it.

### DIFF
--- a/upgrades/2024-upgrade-to-2.36.php
+++ b/upgrades/2024-upgrade-to-2.36.php
@@ -5,7 +5,7 @@
  * It is idempotent i.e. safe to run more than once in Jethro 2.36 and above.
  *
  * Jethro 2.36 introduced the person_status table (https://github.com/tbar0970/jethro-pmm/issues/1035).
- * This script edits reports (person_queries) stored in the database, to reference IDs in person_status.
+ * This script edits reports (person_queries) and action plans (action_plan) stored in the database, to reference IDs in person_status.
  */
 
 ini_set('display_errors', 1);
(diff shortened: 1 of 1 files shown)


jefft:
Also need to fix action plans:

image

tbar0970:
> Also need to fix action plans:


What's the issue with action plans?  If nobody ticked the "set age bracket" checkbox, it doesn't matter that the value field has defaulted to adult.

jefft:
This PR is about statuses, not age brackets. Statuses changed from a string to a foreign key reference. The current `action_plan.actions` serialization for 'Set Status' to 'Regular':

```
{s:6:"status";a:2:{s:5:"value";s:1:"0";s:3:"add";b:0;}
```

The `value` needs to change to an integer. I'll update the PR soon.

tbar0970:
Ah yes, sorry, crossed wires at my end. Correct about references to status in the action plan serialised params.

jefft:
The same `upgrades/2024-upgrade-to-2.36.php` script now upgrades action plans:

```
root@jethro /srv/www/jethro/2.36.0-DEV/app # php upgrades/2024-upgrade-to-2.36.php
Updating report 298
Updating report 337
Updating report 845
Updating action plan 4
Updating action plan 27
Updating action plan 32
Updating action plan 34
Updating action plan 47
Updating action plan 77
Updating action plan 86
Updating action plan 87
Updating action plan 126
Updating action plan 153
done 
```

PR #1097, opened 2024-11-13
Prevent zero-byte /tmp/jethrozipXXXX files being created.

jefft:
Code like this which appends an extension to `tempnam()` is always bad:
```
$zipFilename = tempnam(sys_get_temp_dir(), 'jethrozip').'.zip';
```
because:

1.  `tempnam()` actually creates a file `/tmp/jethrozipXXXX`
2. `$zipFilename` is set to `/tmp/jethrozipXXXX.zip`
3. `$zipFilename` is used
4. `$zipFilename is unlinked to delete it
5. `/tmp/jethrozipXXXX` is still there


jefft:
Actually, this bit of code looks unused. I can't see any way to get a zip of Documents files.

Issue #1104, opened 2024-12-09
Show parents of people in a report


jefft:
Jethro SMS has the useful ability to send to the _parents of_ people in a report, rather than the persons themselves:

![image](https://github.com/user-attachments/assets/a946ace8-ddc4-42e7-871e-502b90260d0b)

This would be a useful capability to add to reports. For instance, find kids in the Youth Group, and then display their parents, so we know who to contact.

tbar0970:
Yes! I was recently thinking the same.
The report condition could be "...who have a family member in [group]"

jefft:
> The report condition could be "...who have a family member in [group]"

That sounds good. Implemented in PR https://github.com/tbar0970/jethro-pmm/pull/1105

PR #1105, opened 2024-12-09
#1104: Let reports filter on "a family member is group X"  

jefft:
allowing us to e.g. find family members of Youth Group members. A "Family Member In Group" column can be added to results to see which particular family member was in the group. 

For example, find all family members of our SALT kids group members, who joined in the last few months. The membership status filter lets us exclude Leaders.

image

We can then ask to see the person in the group:

image

### DIFF
--- a/db_objects/person_query.class.php
+++ b/db_objects/person_query.class.php
@@ -1382,7 +1382,7 @@ class Person_Query extends DB_Object
 						$query['select'][] = 'all_members.names as `All Family Members`';
 						break;
 					case 'familymember_group_members':
-						$query['select'][] = 'familymember_in_required_group.familymember_name as `Family Member In Group`';
+						$query['select'][] = 'GROUP_CONCAT(DISTINCT familymember_in_required_group.familymember_name SEPARATOR ", ")  as `Family Members In Group`';
 						break;
 					case 'adult_members':
 						/*
--- a/db_objects/person_query.class.php
+++ b/db_objects/person_query.class.php
@@ -444,8 +444,55 @@ class Person_Query extends DB_Object
 		</div>
 		<?php
 	}
+?>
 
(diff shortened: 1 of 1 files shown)


tbar0970:
I haven't tested yet, but  I have a suspicion: if a family has 2 kids in the youth group, will both kids' names be shown in the "names of family membes in the group" column ??

tbar0970:
I'm also wondering if the "membership status" and "joined group..." sub-filters for this new feature are not useful enough to justify the extra complexity they add.  Thoughts?

jefft:
> I'm also wondering if the "membership status" and "joined group..." sub-filters for this new feature are not useful enough to justify the extra complexity they add. Thoughts?

I thought the same initially, but changed my mind, and went all-in, for two reasons:

1) limiting to "membership status" is actually critical to my use-case. Our youth group ("SALT") has membership statuses of "Leader", as well as "Regular", "Irregular", etc. I want to get the parents of youth group kids, not the parents of youth group _leaders_! So filtering on "membership status" is necessary.

2) It's a large'ish patch, but it is mostly just copy-and-paste-and-tweak of the "is in group" feature, with little new complexity:
image


Here is a screenshot of the same file (`person_query.class.php`) scrolled to the "is in group" code (left) and the new "is family member in group" code (right). The HTML is identical, except with `_familymember` added to variables:


image

The SQL is likewise very similar:

image

So for the sake of consistency, I didn't monkey with the code I was copying. I left it feature-complete.

jefft:
> I haven't tested yet, but I have a suspicion: if a family has 2 kids in the youth group, will both kids' names be shown in the "names of family membes in the group" column ??

That's a good point. Only one parent showed.

I've amended the PR to show all members. E.g. if both Jimmy and Sally are in youth group, both their names are listed in "Family Members In Group" for each of their parents (Joe and Joanne):

image

Issue #729, opened 2021-10-17
Autosuggest failing with MySQL 8 


2SJH:
With a new installation on Ubuntu 20.04, I've had issues with the rank keyword, but after working through that I'm finding that autosuggest doesn't work. The Javascript console reveals this:

Uncaught SyntaxError: expected expression, got '<'
    setSuggestions http://localhost/jethro/resources/js/bsn_autosuggest.js?t=1634468773:186

This is caused by an uncaught exception in a HTTP request

GET | http://localhost/jethro/?call=find_person_json&search=h

SQLSTATE[42000]: Syntax error or access violation: 1055 Expression #3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'pp.first_name' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

Line 162 of File /home/www/html/jethro/include/jethrodb.php

This can be overcome by turning off the ONLY_FULL_GROUP_BY mode in MySQL

This command will temporarily clear the flag, allowing the query to work:
SET GLOBAL sql_mode=(SELECT REPLACE(@@sql_mode,'ONLY_FULL_GROUP_BY',''));

tbar0970:
Another way of working around this for now is to enable the STRICT_MODE_FIX setting in conf.php
```
// Whether to explicity turn of MySQL's strict mode.
// Enable this if you experience strict-mode-related database errors and can't adjust your central Mysql config.
define('STRICT_MODE_FIX', TRUE);
```

tbar0970:
Thanks for the info.  I've done some reading which is telling me that
> NO_AUTO_CREATE_USER is enabled by default in MySQL 5.7 but not supported in MySQL 8.0.
so as well as the problem you report in this issue, it looks like STRICT_MODE_FIX also needs adjusting.

PR #1129, opened 2025-01-14
Replace UNION with LEFT JOINs to fix query in ONLY_FULL_GROUP_BY mode

radarhere:
Searching for a person (/?call=find_person_json&show-absence-date=&search=test) with Jethro and MySQL 8.4.2, I received an error
> SQLSTATE[42000]: Syntax error or access violation: 1055 Expression 3 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'pp.first_name' which is not functionally dependent on columns in GROUP BY clause; this is incompatible with sql_mode=only_full_group_by

See https://dev.mysql.com/doc/refman/8.4/en/group-by-handling.html
> MySQL implements detection of functional dependence. If the [ONLY_FULL_GROUP_BY](https://dev.mysql.com/doc/refman/8.4/en/sql-mode.html#sqlmode_only_full_group_by) SQL mode is enabled (which it is by default), MySQL rejects queries for which the select list, HAVING condition, or ORDER BY list refer to nonaggregated columns that are neither named in the GROUP BY clause nor are functionally dependent on them.

Looking at the query, I think a simple solution is to remove the GROUP BY altogether - the UNION can be replaced with LEFT JOINs.

### DIFF
--- a/db_objects/person.class.php
+++ b/db_objects/person.class.php
@@ -502,39 +502,31 @@ class Person extends DB_Object
 	{
 		$db = $GLOBALS['db'];
 		$SQL = '
-			SELECT pp.id, pp.*
-			FROM (
-				SELECT p.*
-				FROM person p
-				WHERE (
-					(first_name LIKE '.$db->quote($searchTerm.'%').')
-					OR (last_name LIKE '.$db->quote($searchTerm.'%').')
-					OR (first_name LIKE '.$db->quote('% '.$searchTerm.'%').')
-					OR (last_name LIKE '.$db->quote('% '.$searchTerm.'%').')
-					OR (CONCAT(first_name, " ", last_name) LIKE '.$db->quote($searchTerm.'%').')
+			SELECT p.*
+			FROM person p
+			JOIN person_status ps ON ps.id = p.status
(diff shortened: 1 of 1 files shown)


tbar0970:
Jethro actually relies on the older, more permissive MySQL behaviour in a bunch of places, so this PR wouldn't solve the larger issue. Jethro needs ONLY_FULL_GROUP_BY to be turned off.  See https://github.com/tbar0970/jethro-pmm/issues/729#issuecomment-951536521

radarhere:
Thanks for your review.

Are you saying that if I would like this feature, I should create a larger PR to handle it all at once?
Or are you closing the book on this for now, possibly due to limited time to review changes?

tbar0970:
Yeah, if you're keen for Jethro to be compatible with ONLY_FULL_GROUP_BY, we'd want a PR that covers all the relevant cases. I reckon that would be a lot of work to track them all down, rewrite the queries, and regression test it all.  And yes, review time is limited!

As far as I'm aware there's no sign of ONLY_FULL_GROUP_BY becoming compulsory in MySQL (Jethro's not the only app that relies on the older behaviour), so I don't really see it as worth the effort at this point.

PR #1134, opened 2025-01-19
Updated offset access syntax

radarhere:
When printing envelopes for a family in PHP 8.3, I received various errors of
> PHP Fatal error:  Array and string offset access syntax with curly braces is no longer supported

This fixes that.

### DIFF
--- a/calls/call_envelopes.class.php
+++ b/calls/call_envelopes.class.php
@@ -516,7 +516,7 @@ function GetStringWidth($s)
 	$w=0;
 	$l=strlen($s);
 	for($i=0;$i<$l;$i++)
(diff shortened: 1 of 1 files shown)


tbar0970:
Thanks. Looks like these were from an old library used for envelope generation.  Useful to know that some people still generate envelopes!

radarhere:
Oh, to be clear, I wasn't actually using the feature in a real scenario, just testing the software.

If the code is still present in the repository, then my expectation is that it is helpful to improve it, even if it's not a problem for me personally.

tbar0970:
Yes indeed. Thanks again.

PR #1135, opened 2025-01-19
Declare class property

radarhere:
When adding a family in PHP 8.3, I see in the error log
> Creation of dynamic property View_Families__Add::$_similar_families is deprecated - Line 99 of views/view_2_families__3_add.class.php

This PR declares the class property to remove the warning.


tbar0970:
Thanks!

Issue #21, opened 2015-03-10
Feature request - Attendance report for more than one group


AlwynBarry:
Problem and rationalle:

At the moment one can input and display attendance for more than one congregations, but not more than one group.  So, if you have groups to maintain records of attendance at (for example) many Bible Study Groups, it is not currently possible to input or display these in one go.  But, for tracking attendance over multiple related groups this is a really useful feature ... and the functionality must be basically similar as the implementation for congregations.

Feature Request:
So, the feature request is simple: provide the ability to report attendance over a number of groups at the same time, in the same way as we already have for congregations.

tbar0970:
Done in latest version

PR #1138, opened 2025-01-19
Fix queries in ONLY_FULL_GROUP_BY mode

radarhere:
Addresses #729, building on #1129

I've searched through Jethro for any query containing 'GROUP BY' and run each of them. I found three that caused an error when ONLY_FULL_GROUP_BY mode was enabled. This addresses them, removing the requirement to disable ONLY_FULL_GROUP_BY.

I understand it has been said that [this is not a priority](https://github.com/tbar0970/jethro-pmm/pull/1129#issuecomment-2600404692), but thought I would offer this anyway. It is possible less queries are involved than was suspected.


jefft (commented):
I like this patch, as it fixes some daftness in the old code (using MySQL to calculate a date offset and then returning it for every person record). It doesn't appear to have anything to do with ONLY_FULL_GROUP_BY though?

The PHP-calculated date _should_ be the same as MySQL-calculated, since the timezone is set the same:

https://github.com/tbar0970/jethro-pmm/blob/a5ceb2543065d801c1556a6babe8926d9c7e11cc/include/init.php#L73-L75


It works in practice:

```sh
$ mysql -sNBe "select curdate()  + interval 7 day;"
2025-01-27
$ php -r '$ini = ["REMINDER_OFFSET" => 7]; echo $expiryDate = date("Y-m-d", strtotime($ini["REMINDER_OFFSET"] . " day")) . PHP_EOL;'
2025-01-27
```

radarhere:
The error generated in master by date_reminder.php is
> Syntax error or access violation: 1055 Expression #21 of SELECT list is not in GROUP BY clause and contains nonaggregated column 'jethro.cfv.value_date' which is not functionally dependent on columns in GROUP BY clause

My change removes 'cfv.value_date' altogether.

If you like that part of this PR for non-GROUP BY reasons, I'm perfectly happy to create a separate PR with just that.

sevasek:

PR #1175, opened 2025-02-25
Handle attempt to save duplicate group membership status option

### LINKED ISSUE #1163: Adding a duplicate Group Membership Status throws an error
In Jethro, every group member has a membership 'status', from a list of available Group Membership Status Options configured under Admin -> System Configuration.

If a value is accidentally duplicated, as 'None' is in this example:

Image

then on saving, Jethro blows up with an error:

```
PHP message: SQLSTATE[23000]: Integrity constraint violation: 1062 Duplicate entry 'None' for key 'label' - Line 383 of /srv/www/jethro/2.36.0/app/views/view_10_admin__6_system_configuration.class.php
```

By contrast, duplicating Person Status results in a decent error message:

Image

radarhere:
Resolves #1163


tbar0970:
Question: did you check what happens if only the case is different - "core" vs "Core"? (Possibly the db collation will handle this for us but I’m not sure)

radarhere:
When I test it, it doesn't make a difference - both capitalisations are caught and show the new message that it did "not save new group membership status option".

PR #1176, opened 2025-02-25
Simplify code by determining expiry date in PHP instead of SQL

radarhere:
Cherry-picking a commit from #1138 that received a [positive response.](https://github.com/tbar0970/jethro-pmm/pull/1138#pullrequestreview-2561020379)

Rather than calculating the expiry date in SQL and returning it with each row
https://github.com/tbar0970/jethro-pmm/blob/d752c771097c017bf8e4f3ff0dfc8466d97bf33d/scripts/date_reminder.php#L43
https://github.com/tbar0970/jethro-pmm/blob/d752c771097c017bf8e4f3ff0dfc8466d97bf33d/scripts/date_reminder.php#L65

we could just calculate it in PHP
```php
$expiryDate = date('Y-m-d', strtotime($ini['REMINDER_OFFSET'].' day'));
...
WHERE cfv.value_date = '.$expiryDate.'
```

### DIFF
--- a/scripts/date_reminder.php
+++ b/scripts/date_reminder.php
@@ -40,7 +40,8 @@ $GLOBALS['user_system']->setCLIScript();
 $GLOBALS['system'] = System_Controller::get();
 //error_reporting(E_ALL);
 
-$SQL = 'SELECT p.*, cfv.value_date AS expirydate ';
+$expiryDate = date('Y-m-d', strtotime($ini['REMINDER_OFFSET'].' day'));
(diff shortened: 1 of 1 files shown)


tbar0970:
Looks sensible. Have you tested it on a running date reminder script? I don't have one set up ATM.

radarhere:
Ok, I've tested using a variation of date_reminder_sample.ini. I've made some additional changes, and it works.

PR #1183, opened 2025-02-28
Added PHP lint workflow

radarhere:
Here's an idea - by running the PHP linter on pull requests with GitHub Actions, you could quickly see if there are any syntax errors in a script.

I've started the version matrix at 8.3 just to use glob patterns. See https://php.watch/versions/8.3/cli-lint-multiple-files

### DIFF
--- a/.github/workflows/lint.yml
+++ b/.github/workflows/lint.yml
@@ -1,0 +1,23 @@
+name: Lint
+
+on: [push, pull_request]
+
+jobs:
+  lint:
+    runs-on: ubuntu-latest
+    strategy:
+      fail-fast: false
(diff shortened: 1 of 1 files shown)


tbar0970:
Sounds useful. I haven't used github actions before. Does it block the push to github on error? Or send some kind of notification?

radarhere:
It doesn't block anything, but
- in the 'Actions' tab you will be able to see status of all new commits, including directly pushed changes
- in a pull request you will see status checks at the end, like https://github.com/zlib-ng/zlib-ng/pull/1875. It doesn't stop anyone from merging it, but it serves as a red flag

PR #1187, opened 2025-03-08
Added checkboxes to remove photo from person or family

### LINKED ISSUE #602: delete/remove photos
am I correct in thinking there is no easy way to delete or remove a photo from a family or a person?

radarhere:
Resolves #602

If a photo is present on a family or person, adds a checkbox to 'Photo'.

Screenshot 2025-03-08 at 4 14 02 pm

When checked, it will hide 'Choose file', to clarify that you should only be removing or adding a new photo, not both at once.

Screenshot 2025-03-08 at 4 15 31 pm

### DIFF
--- a/db_objects/family.class.php
+++ b/db_objects/family.class.php
@@ -244,6 +244,14 @@ class family extends db_object
 	function printFieldInterface($name, $prefix='')
 	{
 		if ($name == 'photo') {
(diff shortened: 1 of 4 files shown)


tbar0970:
Thanks for this. I've also added some further interface enhancements I've been wanting to do for a while.

Issue #1243, opened 2025-07-08
Calling /members/?view=_edit_me directly without being logged in breaks Jethro, loses family phone numbers


jefft:
Say the members interface is enabled, and a church member is sent a link directly to `/members/?view=_edit_me`. They are not logged in, so Jethro presents the members login page.

Upon entering their email and password, Jethro will then get confused and breaks:

<img width="977" height="823" alt="Image" src="https://github.com/user-attachments/assets/c1789fa4-f58d-44ad-99d1-626e5f5f8451" />

Unfortunately, at this point, Jethro has also blanked the phone numbers of all members of that person's family. At least it is kind enough to add this to the history:

<img width="680" height="308" alt="Image" src="https://github.com/user-attachments/assets/dc49702c-f457-43c3-8e4c-396cd164f852" />

The stacktrace in 2.36.1:
```
Undefined index: person_form_token
Line 945 of /home/jethro/code/2.36.1/app/db_objects/person.class.php

USER:       1901
REFERER:    https://jethro/members/?view=_edit_me
```
```
Trying to access array offset on value of type null
Line 945 of /home/jethro/code/2.36.1/app/db_objects/person.class.php

USER:       1901
REFERER:    https://jethro/members/?view=_edit_me
```

In 2.36.0:

```
Undefined index: person_form_token
Line 938 of /home/jethro/code/2.36.0/app/db_objects/person.class.php
```
in 2.35.1:
```
Undefined index: person_form_token
Line 934 of /home/jethro/code/2.35.1/app/db_objects/person.class.php
```

tbar0970:
I'd be glad for anyone to check my work here.

jefft:
Confirmed that the original problem is fixed.

PR #1244, opened 2025-07-08
1243 - fix breakage and data corruption when /members/?view=_edit_me invoked without login

jefft:
Per #1243 description. 

What happens is:

Our user requests `/members/?view=edit_me`. This invokes `/members/index.php`, the last part of which is:

https://github.com/tbar0970/jethro-pmm/blob/7cf61545f5769e77ae8573fb3fd6395f704680b2/members/index.php#L50-L59

Because the user is not logged in, `Member_User_System` initially prints the login screen and exits:

https://github.com/tbar0970/jethro-pmm/blob/7cf61545f5769e77ae8573fb3fd6395f704680b2/members/include/member_user_system.class.php#L56-L58

the user submits the login form. Now `Member_User_System` is happy - it processes the `$_POST` and logs the user in. Now it's `System_Controller`'s turn. It instantiates `View__Edit_Me` and calls `processView()`.

At this point it all goes wrong. We're processing a view, but the only `$_POST` fields are `username` and `password`. Yet `View__Edit_Me` doesn't know this. It thinks that, purely because it's POST operation, it can expect to find all the params form the edit_me form:

https://github.com/tbar0970/jethro-pmm/blob/7cf61545f5769e77ae8573fb3fd6395f704680b2/members/views/view_0_edit_me.class.php#L42

Everything thing breaks, as none of the `edit_me` form fields are are found.

--- 
The fundamental problem here is that Jethro tries to handle authentication and view logic all at once.

Most webapps have a `/login` URL. If one requests `/edit_me` without being logged in, the app will redirect to `/login?redirect=edit_me`. Then `/login` runs, and in a completely separate HTTP request, `/edit_me` runs. There is no mixing of auth with view logic.

Jethro doesn't have a separate `/login`. When `/members/?view=edit_me` is called, both authentication and view handling happens at the bottom of `members/index.php`, by `Member_User_System` `System_Controller` respectively:

https://github.com/tbar0970/jethro-pmm/blob/7cf61545f5769e77ae8573fb3fd6395f704680b2/members/index.php#L50-L59

This can be made to work, if each data-changing view (like edit pages) validates `$_POST` contains the view's form fields, not the login form's fields. This members `edit_me` is the only instance I've found where only `$_POST` is needed:

https://github.com/tbar0970/jethro-pmm/blob/7cf61545f5769e77ae8573fb3fd6395f704680b2/members/views/view_0_edit_me.class.php#L42

So my suggested fix is to check that `$_POST` contains person fields - and if not, treat it like a GET and print the form without processing.


members/views/view_0_edit_me.class.php
```diff
@@ -125,4 +125,23 @@ private function isMe($person)
...
 
+    /**
+     * Examine $_POST to see if it contains what we expect from the edit_me form being submitted. This will not be true if the initial request was unauthenticated and User_System served up the login form instead - then $_POST will contain just 'email' and 'password', and we can't proceed. See #1243.
+     * @return bool Whether to process the form as if submitted
+     */
+    public function EditMeFormSubmitted(): bool
+    {
+        if (empty($_POST)) return false;
+
+        // This is pretty ugly. We have no single parameter we can check that we (edit_me) definitely set, but we will has a bunch of person_* params, each of which will has a csrf token. So if we see any of those, proceed.
```
radarhere, line 136:
```suggestion
        // This is pretty ugly. We have no single parameter we can check that we (edit_me) definitely set, but we will have a bunch of person_* params, each of which will have a csrf token. So if we see any of those, proceed.
```

jefft:
Thanks @radarhere, changed merged.

PR #1266, opened 2025-08-29
If an entered field is an int, validate it and return a user-level error

### LINKED ISSUE #1265: Component library importer breaks if CSV contains invalid numbers for Length_Mins or CCLI_Number
When importing songs from CSV, there is a mandatory 'Length_Mins' column. Say we enter '4:55' meaning '4 minutes, 55 seconds', mistakenly thinking this is valid. Jethro breaks, because internally it wants an integer rounded-up minute. Jethro ought to return a CSV parse error, rather than explode.

Same breakage if one enters a non-numeric character in `CCLI_Number`.

To replicate:
 - Locally, create `/tmp/song.csv` containing:
```csv
Title,Alt_Title,Length_Mins,Runsheet_Title_Format,Handout_Title_Format,Show_in_handout,CCLI_Number,Credits,Content
My Song,,4:55,,,,,,,
```
 - Go to Services -> Component Library
 - Click 'Import'
 - Select `song.csv`, any congregation, and import

Jethro breaks with 'An error occurred', and logs:

```
SQLSTATE[01000]: Warning: 1265 Data truncated for column 'length_mins' at row 1 - Line 240 of /home/jethro/code/2.36.1/app/include/db_object.class.php
```

jefft:
Fixes #1265

The validation happens way down in the DB_Object class. That is also where `allow_empty` and `max_length` constraints are checked, so why not basic type checking too.

Perhaps other types besides 'int' should be validated in the same place, like 'datetime'. However I don't know how to trigger saving of a datetime. Without knowing how to trigger that codepath, I wouldn't want to risk breaking stuff. So validation is for int only.

### DIFF
--- a/include/db_object.class.php
+++ b/include/db_object.class.php
@@ -549,12 +549,9 @@ class db_object
 		if ($this->fields[$name]['type'] == 'int') {
 			if (!array_get($this->fields[$name], 'allow_empty', true) || ($value !== '')) {
 				$strval = (string)$value;
-				for ($i=0; $i < strlen($strval); $i++) {
-					$char = $strval[$i];
-					if ((int)$char != $char) {
-						trigger_error(ents($value).' is not a valid value for integer field "'.$name.'" and has not been set', E_USER_NOTICE);
-						return;
-					}
+				if (!filter_var($strval, FILTER_VALIDATE_INT)) {
+					trigger_error(ents($value).' is not a valid value for integer field "'.$name.'" and has not been set', E_USER_NOTICE);
+					return;
 				}
 			}
 		}


include/db_object.class.php
```diff
@@ -612,6 +612,17 @@ public function validateFields()
 				trigger_error('The value for '.array_get($details, 'label', $id).' is not the correct length (must be exactly '.$details['fixed_length'].' characters)', E_USER_NOTICE);
 				$res = FALSE;
 			}
+			if ($res) { // If not already errored out..
```
radarhere, line 615:
```suggestion
			if ($res) { // If not already errored out...
```

jefft:
Actually, this breaks import of previously allowed CSV:

```
Title,Length_Mins,CCLI_Number
My Song2,0,,
```

CCLI_Number is not required, but now the line breaks with 'CCLI Number value "" is not a valid integer'

I'll work on a fix.

jefft:
I noticed that, without this patch, Jethro breaks under PHP 7.4 but correctly catches the CSV error when running PHP 8.4:

image

The int-checking code is:

https://github.com/tbar0970/jethro-pmm/blob/a41fca1131f05d7ec8d3e24b7821a1caf8c5aae9/include/db_object.class.php#L549-L560

This convoluted "is this string an int" code doesn't work properly, because PHP is an insane language where `0 == ':'` evaluates to `true`:

image

at least until PHP 8 implemented [PHP RFC: Saner string to number comparisons](https://wiki.php.net/rfc/string_to_number_comparison), and changed the evaluation to false:

image

I have changed my patch to use `filter_var($strval, FILTER_VALIDATE_INT)`, which seems to work in PHP 7.4 and 8.x:

image

jefft:
I've just force-pushed a fix - "0" was being classed as invalid.

PR #1273, opened 2025-08-30
Correctly identify whether a service has runsheet items. 

### LINKED ISSUE #1272: Service Schedule icons suggest a runsheet does not exist when it does
Services in Jethro's list of services (Services -> List All) have an icon denoting whether that service has a runsheet:
 - ⊕ indicates no runsheet exists yet - click the (⊕ to add one)
 - ☰ indicates a runsheet exists

For example, [here](https://easyjethro.com.au/demo//?view=services__list_all&congregations%5B%5D=2&congregations%5B%5D=3&congregations%5B%5D=4&start_date_d=11&start_date_m=1&start_date_y=2026&end_date_d=11&end_date_m=1&end_date_y=2026) the 4pm service apparently has no runsheet, and the 6pm service does:

Image

But that's not true. Clicking the 4pm service shows it does have a runsheet: 

Image

Jethro is erroneously showing "⊕" (no runsheet) for services that actually have a runsheet - in particular, on runsheets that have been edited (or copied from an edited runsheet).

jefft:
Fixes #1272

The code's task here is to identify whether a `service` has any associated `service_item`'s.

The old code did this in SQL:
```sql
SELECT
...
   IF (si.id IS NULL, 0, 1) as has_items
FROM service
   LEFT JOIN service_item si ON si.serviceid = service.id AND si.`rank` = 0`
```
The logic is: "join `service` on the first (`rank=0`) `service_item`. If that exists, we've got at least one runsheet item, so `has_items` is true"

But because service_items can be re-ranked, the first item may not have `rank=0`. It may be `rank=1`.

It's better to ignore `rank` and just check for the existence of any associated `service_item`. That's what my fix does.

### DIFF
--- a/db_objects/service.class.php
+++ b/db_objects/service.class.php
@@ -469,8 +469,7 @@ class service extends db_object
 		$res = parent::getInstancesQueryComps($params, $logic, $order);
 		$res['select'][] = 'GROUP_CONCAT(CONCAT(sbr.bible_ref, "=", sbr.to_read, "=", sbr.to_preach) ORDER BY sbr.order_num SEPARATOR ";") as readings';
 		$res['from'] .= ' LEFT JOIN service_bible_reading sbr ON service.id = sbr.service_id';
-		$res['select'][] = 'IF (si.id IS NULL, 0, 1) as has_items';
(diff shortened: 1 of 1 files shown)


tbar0970:
Thanks. I decided to fix it differently. Your subquery works, but it's more efficient to avoid a correlated subquery.

PR #1274, opened 2025-08-30
Fix bug where ad-hoc runsheet items weren't copied by 'Copy from previous'

### LINKED ISSUE #1271: Service runsheet 'Copy from previous' does not copy ad-hoc items
### Background

Jethro lets you define runsheets for each service (Services -> List All, then click + on a service+date). For a new week, the runsheet is initially blank, and there is a handy 'Copy from previous' button:

Image

this lets you pick what to copy:

Image

### The bug

Jethro runsheets can contain:
 -  predefined items, like songs, from the service component library
 -  "ad hoc" items specific to the runsheet.

When 'Copy from previous' is clicked "ad-hoc" items are not copied. Only 'service component' items are copied (if their category is checked in the list).

Some churches have almost the entire runsheet composed of ad-hoc items. 'Copy from previous' is thus useless.  I'd guess most churches have at least some ad-hoc items, and the loss of these in runsheet copy would be surprising.

### Correct behaviour

I suggest ad-hoc items be copied by default. This seems good/expected default behaviour, and they can always be removed manually later.
### Replication

 - Pick any existing service runsheet
 - Click an item and 'Add ad-hoc item below':
Image
 - Add an ad-hoc item with a distinctive title
Image

Image

- Click 'Save'
- Click 'Back to service schedule'
- For the same congregation (column), find a week without a runsheet, and click the + icon to add one:

    Image

    (if you find that, surprisingly, a runsheet already exists, that's bug #1272. Just delete the runsheet items)

 - Click 'Copy from previous', and pick your previously edited service:

    Image

Your newly populated runsheet will not have the ad-hoc item added:

Image

jefft:
Fixes #1271

As a design choice: I initially considered adding a new "Ad-hoc Items" to the list of "things to copy":

image

But this isn't possible easily. The 'Items to copy' list is generated by `print_widget`, which only prints referenced service component categories, and won't let me inject "Ad-hoc Items":

https://github.com/tbar0970/jethro-pmm/blob/c0b3bc2efe75d0a1735188d7b5963308121e9d15/views/view_8_services.class.php#L435-L440

So copying ad-hoc items by default is the pragmatic choice. In the code that's done by _not_ excluding items with a null `categoryid`.

### DIFF
--- a/views/view_8_services.class.php
+++ b/views/view_8_services.class.php
@@ -59,7 +59,7 @@ class View_services extends View
 								}
 							}
 
-							if (!in_array($v['categoryid'], $_REQUEST['copy_category_ids'])) {
+							if (isset($v['categoryid']) && !in_array($v['categoryid'], $_REQUEST['copy_category_ids'])) {
 								unset($newItems[$k]);
 							}
 						}


tbar0970:
I've implemented your original idea of adding it as an option to choose.
Ad hoc items are frequently used for one-off non-repeated items. 
So if I'm accustomed to copying just announcements etc, having the ad hocs come over without being invited would be annoying.

jefft:
Sure, thanks!

Issue #1270, opened 2025-08-30
Allow %alt_title% in service component templates


jefft:
In Jethro, service components have a Title, and then an Alt Title.

One use of Alt Title might be to record the Australian Hymn Book identifier: 

<img width="325" height="231" alt="Image" src="https://github.com/user-attachments/assets/7e27fdde-bda7-4c5d-ae44-1010ac3d989a" />

In runsheets and handouts, it would then be nice to print the Alt Title alongside the Title, by amending the template:

<img width="353" height="472" alt="Image" src="https://github.com/user-attachments/assets/7634163d-a63a-4d9d-9f16-fa5eb5ceb950" />

Unfortunately right now, only `%title%` is substituted, not `%alt_title%`.

PR #1280, opened 2025-09-02
Avoid deprecation warnings in PHP 8.4 introduced in code for #1270

jefft:
My PR #1275 fixing #1270 introduced deprecation warnings under PHP 8.4:

```
str_replace(): Passing null to parameter #2 ($replace) of type array|string is 
deprecated - Line 502 of /home/jethro/code/latest/app/db_objects/service.class.php
```
because alt_title and ccli_number may be null.  If they are null, we want that to be equivalent to "" - hence the null coalescing operator (supported in 7.0).

### DIFF
--- a/db_objects/service.class.php
+++ b/db_objects/service.class.php
@@ -499,8 +499,8 @@ class service extends db_object
 	public function replaceItemKeywords($title, $iteminfo)
 	{
 		$title = str_replace('%title%', $iteminfo['title'], $title);
-		$title = str_replace('%alt_title%', $iteminfo['alt_title'], $title);
-		$title = str_replace('%ccli_number%', $iteminfo['ccli_number'], $title);
+        $title = str_replace('%alt_title%', $iteminfo['alt_title'] ?? "", $title);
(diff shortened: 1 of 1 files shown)


db_objects/service.class.php
```diff
@@ -499,8 +499,8 @@ function getInstancesData($params, $logic='OR', $order='')
 	public function replaceItemKeywords($title, $iteminfo)
 	{
 		$title = str_replace('%title%', $iteminfo['title'], $title);
-		$title = str_replace('%alt_title%', $iteminfo['alt_title'], $title);
-		$title = str_replace('%ccli_number%', $iteminfo['ccli_number'], $title);
+        $title = str_replace('%alt_title%', $iteminfo['alt_title'] ?? "", $title);
+        $title = str_replace('%ccli_number%', $iteminfo['ccli_number'] ?? "", $title);
```
radarhere, line 503:
```suggestion
        $title = str_replace('%alt_title%', $iteminfo['alt_title'] ?? '', $title);
        $title = str_replace('%ccli_number%', $iteminfo['ccli_number'] ?? '', $title);
```

Feel free to ignore this suggestion, it just seems nicer to me to be consistent about quotes within the same line.

PR #1283, opened 2025-09-02
If a song already exists, update it rather than creating a duplicate.

### LINKED ISSUE #1282: FR: Update songs from CSV if Title matches
When importing songs from CSV, often one notices mistakes or omissions in the CSV after importing, and would like to try again, updating Jethro's record of the song to the newest CSV.

This isn't a problem if your songs have a CCLI number. Jethro will (by default) update it's song record to match CSV if the CCLI number matches:

Image

But what about songs that don't have a CCLI? What if `CCLI_Number` is one of the fields you forgot to set initially, and want to update?  Then your import will result in duplicate songs - which then can't even be deleted (#906).

It would be nice if Jethro could also match on song Title, e.g. with a new checkbox:

Image

jefft:
Fixes #1282

image

Title is not, in general, reliable at distinguishing songs, e.g. there are multiple "Hosanna" songs with different CCLIs. So the lookup-by-CCLI is done first, then (if CCLI didn't match) lookup-by-title.

I also want to support having two songs with identical Title, but different Alt_Title. Hence, there's code that only identifies a duplicate if Title and Alt_Title match (if Jethro's record has an Alt_Title), but if Jethro knows of no Alt_Title, then just a Title match is sufficient.

I started by cloning the "does CCLI match" code, but then it started to get really ugly, so broke the new code out into a method, and then, for consistency, did the same for the CCLI matching code.

As a final improvement, the success message now distinguishes the number of songs _created_ vs. those _updated_. For repeated imports, one would want 0 songs created:

image

### DIFF
--- a/db_objects/service_component.class.php
+++ b/db_objects/service_component.class.php
@@ -208,6 +208,14 @@ class Service_Component extends db_object
 		return $res;
 	}
 
+	public static function getAllByTitle()
+	{
+		$SQL = "SELECT title, alt_title, id
+				FROM service_component";
+		$res = $GLOBALS['db']->queryAll($SQL, null, null, true, false);
+		return $res;
+	}
+
 	protected function _printSummaryRows()
 	{
 		$oldFields = $this->fields;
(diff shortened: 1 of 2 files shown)


jefft:
Thanks.  I discovered some more bugs overnight which I'll follow up with in a separate PR.

PR #1287, opened 2025-09-03
Support setting tags on songs imported from CSV

### LINKED ISSUE #1262: Component Library importer should support setting tags
In Services -> Component Library -> Import, it is possible to bulk import 'components' such as songs.

Currently the CSV importer cannot set Tags on imported items. It would be nice if one could have a Tags' column in CSV:

ImageUnder

and have it set a song's tags, just like manual editing would do:

Image

jefft:
Fixes #1262.  Adds support for setting song tags from CSV, via new column(s).

## Code notes

In the code, I needed to create `Service_Component_Tag`s and `Service_Component_Tagging`s only if they didn't already exist. So I copied the `createIfNew` pattern I found in `Person_Note`.

`createIfNew()` is a candidate for being moved up to DB_Object. For that to work, there would need to be a way for each object to tell it "which of my columns are unique". Possibly we could use `_getUniqueKeys()` (currently only used for FK creation). Anyway, this is a bigger refactoring I didn't tackle.

### Sample file

For basic testing, go to Services -> Component Library -> Import, and download the Sample file:

image

The imported songs are now tagged:

image
image

### Detailed testing

This song CSV illustrates setting 0, 1 or 2 tags via multiple columns:

```csv
Title,Tag 1,Tag 2
Tagged Test Song 1,TagOne,
Untagged Test Song 2,,
Tagged Test Song 3,TagTwo,
Tagged Test Song 4,TagOne,TagTwo
```
image

Imported songs are tagged as expected:
image
image
image
image


jefft:
Replaced by https://github.com/tbar0970/jethro-pmm/pull/1341

PR #1293, opened 2025-09-05
Get song search working again, after CCLI search broke it.

### LINKED ISSUE #1292: Service component filtering not working
Go to service components page
search for "jungle"
Lots of non-jungle-related songs come up.

Maybe just in 2.37-RC ?

jefft:
Fixes #1292.

MySQL is the same clown show as PHP when it comes to type coercion. My code added `ccli_number = '$searchterm'` to the SQL. When comparing an int to a string, MySQL typecasts the string to an int, so 'jungle' becomes 0, so all songs with 0 CCLI match.

The fix is to cast `ccli_number` to characters (`CHAR`) before testing.

### DIFF
--- a/db_objects/service_component.class.php
+++ b/db_objects/service_component.class.php
@@ -194,7 +194,12 @@ class Service_Component extends db_object
 		}
 		if ($keyword) {
 			$qk = $GLOBALS['db']->quote("%{$keyword}%");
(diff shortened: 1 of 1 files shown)


tbar0970:
Hmm.  If $keyword isn't a positive integer, wouldn't it be best to omit CCLI-number-matching from the query altogether?

jefft:
Now including a second commit 77f69f56f626ca49e6ae9fe6420053b7aaf540ee that only includes the SQL clause if the search keyword is an integer.  I'm not sure the increased code complexity was worth it. Oh well.

PR #1304, opened 2025-09-12
Service navigation links

### LINKED ISSUE #884: Navigating between services in run sheet editor
Sometimes I'm planning one aspect for many services in one go, eg songs for a month.
When I'm viewing a run sheet it would be handy to be able to quickly go to the next/previous service for that congregation.
eg if I'm viewing 8th March run sheet, a link  to go direct to 1st March or 15th March.

Curently we have a "back to service schedule" link at the top right. This could be somehow extended to include "Go to next/previous 9am service"

jefft:
Fixes #884.

This patch adds a set of navigation links when viewing or editing services:

image

The nav links let one move to:
 - the service schedule (icon replacing old 'Back to service schedule' link)
 - an earlier congregation (left arrow)
 - earlier week (up arrow)
 - later week (right arrow)
 - later congregation (right arrow)

This mirrors the layout in the service schedule.

I initially had text descriptions next to the buttons, but it got really busy and distracting, so I went with all-buttons with rollover help text.

The icon goes grey/unclickable if there is no earlier/earlier congregation, or no further services:

[Screencast_20250912_122115.webm](https://github.com/user-attachments/assets/8db4cd0b-13ba-4c03-ad3c-cda8234622ab)

One annoyance, visible in the video, is that service widths often jump about - some full width, some part-width. That's a separate UI bug.

### Code notes

This PR depends on / incorporates https://github.com/tbar0970/jethro-pmm/pull/1300 which introduced an `action` URL param, which this PR makes use of.

The nav links all go back to the original view, giving it an opportunity to clean up the edit lock, if the page was in edit mode.

### DIFF
--- a/views/view_8_services.class.php
+++ b/views/view_8_services.class.php
@@ -42,65 +42,98 @@ class View_services extends View
 						trigger_error("Your lock expired and your changes could not be saved");
 						return;
 					}
+				}
 
-					if (!empty($_REQUEST['action'])) {
-						$action = $_REQUEST['action'];
-						switch ($action) {
-							case "copy":
-								if (!empty($_REQUEST['copy_service_id'])) {
-									$fromService = new Service((int)$_REQUEST['copy_service_id']);
-									if (!$fromService) {
-										trigger_error("Service ".(int)$_REQUEST['copy_service_id']." not found - could not copy");
-										return;
-									}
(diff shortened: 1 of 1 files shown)


tbar0970:
Looks great. Thanks.

PR #1311, opened 2025-09-17
Remove upgrade_session_cookie() function

jefft:
Jethro has this odd behaviour where it sets the `JethroSess` cookie, then immediately deletes it and re-issues it in `upgrade_session_cookie`.  The latter appears to be a later addition, to add `SameSite=Lax` to previously issued cookies.

This PR removes `upgrade_session_cookie` and sets `JethroSess` just once. In PHP 7.3+ (ref. https://github.com/tbar0970/jethro-pmm/pull/1310) we can set SameSite directly in session_set_cookie_params(), and it's even the default in PHP 8.x.

It is also unnecessary to call session_set_cookie_params() before session_regenerate_id(true), as the new cookie already inherits the set cookie params, including SameSite.


include/init.php
```diff
@@ -61,13 +61,18 @@
...
 }
 
-// If max length is set, set the cookie timeout - this will allow sessions to outlast browser invocations
-$expiryTime = defined('SESSION_MAXLENGTH_MINS') ? SESSION_MAXLENGTH_MINS * 60 : NULL;
-session_set_cookie_params($expiryTime, parse_url(BASE_URL, PHP_URL_PATH));
 if (session_id() == '') {
+  	// If max length is set, set the cookie timeout - this will allow sessions to outlast browser invocations
+  	$expiryTime = defined('SESSION_MAXLENGTH_MINS') ? SESSION_MAXLENGTH_MINS * 60 : NULL;
+  	session_set_cookie_params([
+  		'lifetime'=> $expiryTime,
```
radarhere, line 68:
```suggestion
  		'lifetime' => $expiryTime,
```

PR #1312, opened 2025-09-17
Error handling: replace alert() messages with inline error text for form validation

jefft:
Successor to PR 1308 that got too messy..

Jethro has old-school error reporting: Javascript alerts:

image

For forms, it would be better to highlight the missing/incorrect form field value, with the error message below it.

This PR attempts to fix this. E.g. on /members login, rather than an alert:

image

# Testing

### Email

Email fields were not previously validated, because the code (in general.php) did not add a `valid-email` class, because (I think) it tries to handle text and email in one codepath.

Anyway, emails are now validated.

[Screencast_20250917_205920.webm](https://github.com/user-attachments/assets/4b752db1-6a4c-4549-9d31-1272c1697471)

### Bible refs

E.g. at Services -> List All -> Edit Schedule

[Screencast_20250917_210454.webm](https://github.com/user-attachments/assets/a3d26d23-645e-48e3-a3e3-7681c846e3b5)

### Dates (day, year fields)

This shows how a single 'control group' can have two kinds of error message (for day and year):

[Screencast_20250917_210606.webm](https://github.com/user-attachments/assets/dfe12456-75eb-4a78-a03f-87e8d0839c7e)

### Phone numbers

[Screencast_20250917_211703.webm](https://github.com/user-attachments/assets/731e935a-205f-4caa-81a8-e8923c295157)

Also, phone numbers were previously only validated on submit, not on field blur. 

### Regex-validated inputs

E.g. a WWCC person field configured with regex:

image

[Screencast_20250917_212855.webm](https://github.com/user-attachments/assets/3208d1ad-c650-4fc4-88b2-5d1514e19c52)

Note, HTML has its own regex input validation these days, via a `pattern` attribute. I used it too, just out of curiousity. It makes the text go red until the regex matches. 

### `data-require-fields`

Jethro has a `data-require-fields` attribute, which can be set on a submit button. It allows one to dynamically make an input required. Currently this is only used in person bulk actions.

Validation only takes place when the button is pressed:

[Screencast_20250917_213601.webm](https://github.com/user-attachments/assets/caf0d378-365a-4309-b29a-5d480f86c3a7)

### DIFF
--- a/include/general.php
+++ b/include/general.php
@@ -279,6 +279,7 @@ function print_widget($name, $params, $value)
 				><?php echo ents($value); ?></textarea>
 				<?php
 			} else {
+				if ($params['type']=='email') $classes .= " valid-email";
 				$width_exp = empty($params['width']) ? '' : 'size="'.$params['width'].'"';
 				$regex_exp = empty($params['regex']) ? '' : 'regex="'.ents(trim($params['regex'], '/ ')).'"';
 				$placeholder_exp = empty($params['placeholder']) ? '' : 'placeholder="'.ents($params['placeholder']).'"';
--- a/include/general.php
+++ b/include/general.php
@@ -281,7 +281,7 @@ function print_widget($name, $params, $value)
 			} else {
 				if ($params['type']=='email') $classes .= " valid-email";
 				$width_exp = empty($params['width']) ? '' : 'size="'.$params['width'].'"';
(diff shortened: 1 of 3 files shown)


jefft:
Replaced by https://github.com/tbar0970/jethro-pmm/pull/1349

PR #1345, opened 2025-10-28
New DB_DSN config setting allowing arbitrary database configs, including unix socket auth.

### LINKED ISSUE #1336: Support for connecting to database with unix sockets
Jethro currently connects to MySQL over TCP, with a username and password.  If PHP is running as a dedicated user (`jethro`), then it would make sense to allow Jethro to connect to MySQL via unix socket.

```sql
-- In MySQL
alter user 'jethro'@'localhost' identified with unix_socket;
-- Alternatively:
-- create user if not exists 'jethro'@'localhost' identified with unix_socket;
```

This will be [faster](https://www.percona.com/blog/need-to-connect-to-a-local-mysql-server-use-unix-domain-socket/) than a localhost TCP connection, but the main advantage is operational. Restoring a Jethro backup in another MySQL server, one doesn't have to extract the DB_PASSWORD from conf.php and then run:
```sql
create user if not exists 'jethro'@'localhost' identified by '<PASSWORD FROM conf.php>';
```
With socket auth, there is no password needed:
```sql
create user if not exists 'jethro'@'localhost' identified with unix_socket;
```
and as a benefit, `sudo -u jethro mysql` works too.

jefft:
Fixes #1336


jefft:
> Can we add this setting to conf.php.sample with a note something like "advanced system admins can use this instead of the database details above" ?

Sure, I've added a `conf.php.sample` comment, focusing on the unix socket auth use-case:
```php
// If your Jethro PHP-FPM runs as a dedicated OS user ('jethro'), you can also enable socket-based authentication in MySQL:
//   create user if not exists 'jethro'@'localhost' identified with unix_socket;
// ..and use it by specifying DB_DSN instead of DB_PASSWORD:
// define('DB_USERNAME', 'jethro');
// define('DB_DSN', "mysql:dbname=jethro");
```

PR #1332, opened 2025-10-11
Fix roster role link in members, and remove role description top-level link in public.

jefft:
Fixes #1330 and #1331 (changes are intedependent).

#1330 is a simple fix:
 - make the roster role's hyperlink page-relative. I.e. change `BASE_URL.'/public/?view=display_role_description...` to `?view=_roster_role_description` in `roster_view.class`.
 - create a new view, [members/views/view_0_roster_role_description.class.php](https://github.com/tbar0970/jethro-pmm/compare/master...jefft:jethro-pmm:push-1330_1331_roster_roles_in_public?expand=1#diff-2cc11bf5c2b9b02c688edd59fd674fc17ee929d9a697b8922b058e3b0203cc85), identical to `public/` equivalent, for the now-relative hyperlink to point to.

#1331 is also a simple but messy business of renaming files and links.

The result:

image

PR #1346, opened 2025-10-28
Fix roster role link in members, and remove role description top-level link in public.

jefft:
Replacement for PR https://github.com/tbar0970/jethro-pmm/pull/1332 - see discussion there.


tbar0970:
I'm going to merge this, and we can finiesse the UI in due course.

Issue #1224, opened 2025-05-31
Make member file sharing more discoverable, and MEMBER_FILES_DIRS user-configurable


jefft:
Jethro has a Documents section, intended mainly for staff files.

It's possible to also share some Documents with 'members' (non-staff, logging in through the /members interface), by setting the `MEMBER_FILES_DIRS` directory to one below the root document path:

```
define('DOCUMENTS_ROOT_PATH', '/home/jethro/home/documents');
define('MEMBER_FILES_DIRS',   '/home/jethro/home/documents/Member_Files');
```

Now, _if one knows about this setting_, one can create a `Member_Files` top-level folder, and share files by putting things in it:

![Image](https://github.com/user-attachments/assets/d525b3f5-8534-4aca-804d-cf443db53482)

The problem is that this feature is undiscoverable without knowing about that `conf.php` setting, and knowing to manually create the folder.

Some documentation would help, but I'm wondering why this setting doesn't appear in the 'Members' section of System Configuration?

Perhaps we could add it? E.g.:

![Image](https://github.com/user-attachments/assets/2f83c139-0a4d-42ef-8b9a-2602333cc613)

tbar0970:
Fair point about discoverability.

The logic of not having that setting on the system config page was security based. Don't really want non-server-admins deciding what directories are exposed to low-privileged jethro users. I don't think there is anything restricting the member docs folder to being a child of the general docs folder. That is perhaps a restriction that could be added.

jefft:
> I don't think there is anything restricting the member docs folder to being a child of the general docs folder. That is perhaps a restriction that could be added.

On the contrary, why should `MEMBER_FILES_DIRS` be anything but a subdirectory of `DOCUMENTS_ROOT_PATH`?

Surely the main use-case is to allow Jethro staff to share certain documents with members?  If so, staff can only access below `DOCUMENTS_ROOT_PATH`, and it makes sense for `MEMBER_FILES_DIRS` to be a subset of that.

Then `MEMBER_FILES_DIRS` can be restricted to be a relative path, and it's no longer particularly security-sensitive.

PR #1347, opened 2025-10-28
Config setting to safely allow Members-visible folders

jefft:
Fixes https://github.com/tbar0970/jethro-pmm/issues/1224. Successor to https://github.com/tbar0970/jethro-pmm/pull/1329

### DIFF
--- a/include/general.php
+++ b/include/general.php
@@ -1079,7 +1079,7 @@ function error_response(int $code, string $message): void {
 
 /** Return $path if it is an existing subdirectory of $base, or false otherwise.
  * @param $base Directory we want to be relative to. May not contain '..'.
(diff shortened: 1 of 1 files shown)


include/general.php
```diff
@@ -1076,3 +1076,33 @@ function error_response(int $code, string $message): void {
 	echo ents($message);
 	exit;
 }
+
+/** Return $path if it is an existing subdirectory of $base, or false otherwise.
+ * @param $base Directory we want to be relative to. May not contain '..'.
+ * @param $path Path to check. E.g. 'foo' returns 'foo', '/foo' returns 'foo'. '' return false. '../' returns false.
```
radarhere:
```suggestion
 * @param $path Path to check. E.g. 'foo' returns 'foo', '/foo' returns 'foo'. '' returns false. '../' returns false.
```

PR #1377, opened 2026-02-09
Remove unused DTMF dialtone generator code.

jefft:
It's pretty neat but.. completely unrelated to Jethro?


tbar0970:
Ha. At the very start, Jethro had a "dial this number" feature - you could hold your landline phone up to the computer speakers and it would connect the call. When i first demoed Jethro, my minister thought that was fantastic, James Bond stuff. Sadly, analog landlines are no more, so the link was removed. Quite right to now remove the library too. Did you hear about the recent security vulnerability found in windows 11, in the fax modem driver from circa 1993?!

PR #1380, opened 2026-02-16
Allow teams to be assigned to rosters

radarhere:
When allocating people to rosters, rather than repeatedly assigning the same people who operate as a unit (e.g. band) to a role, it may be helpful to be able to allocate a group.

Roster roles select people from a group however, and groups don't have subgroups. So instead, I suggest checking all other groups in that same category, and if all of that group's members are in the role group, then it is possible to assign them - this isn't anything that the user isn't already able to do, just slower.

John Smith from the group may be away for a given week though, so once the submission is made, I suggest showing the rostered people as individuals once more.

This would only apply to roles that allow for 'Assign Multiple'.

https://github.com/user-attachments/assets/741e0e4a-0cd7-47e2-898b-d728d349dbdd

### DIFF
--- a/db_objects/roster_role.class.php
+++ b/db_objects/roster_role.class.php
@@ -135,7 +135,7 @@ class Roster_Role extends db_object
 				'order_by' => 'name',
 				'allow_empty' => true,
 				'allow_multiple' => true,
-				'note' => 'Groups whose members can be assigned to this role as a set, eg. \'Band 1\''
+				'note' => 'Groups whose members can be assigned to this role as a batch, eg. \'Band 1\''
 			]
 		] + $this->fields;
 		parent::printForm($prefix, $fields);


tbar0970:
Yeah, interesting, not a bad approach in principle.   But I think we'd need to work on the 'affordance'. As it stands I don't think anyone would spot or understand the feature without it being explained externally.  Is there a way to make it clear that you can create teams by creating a sibling group with a subset of the volunteers?  

Perhaps at the bottom of the volunteer dropdown box could be an entry labelled "Create team....". It opens a modal where you can select members of the volunteer group to add to a team with name XYZ (which is actually a sibling group as per your existing design).  Once that process is done, the volunteer dropdown has "Team: XYZ" available at the bottom.

But some further edge cases would still need attention. If I delete Joe Bloggs from the overall musicians volunteer group, but don't delete him from the team group, then the team group will no longer be regarded as a team, but users won't understand why...

Perhaps it's necessary to actually create the concept of a sub-group, so that Jethro can make sure members of the sub-group are always members of the parent group.

db_objects/person_group.class.php
```diff
@@ -176,6 +176,49 @@ function getMembers($params=Array(), $order_by=NULL)
 		$comps['select'][] = 'pgm.created as joined_group';
 		return $person->_getInstancesData($comps);
 	}
+	
+	function getPossibleSubgroups($member_ids)
+	{
+		$db =& $GLOBALS['db'];
+		$sql = 'SELECT pgm.personid, pgm.groupid, pg.name FROM person_group_membership pgm
+			JOIN _person_group pg ON pg.id = pgm.groupid
+			WHERE pgm.groupid != '.$this->id.' AND pg.categoryid';
```
tbar0970: Even though it's coming from a relatively safe source, please escape all variables passed into a SQL query, either with $db->quote() or by casting as an int. Also below.

radarhere:
Ok, I've changed this to actually use subgroups, instead of just groups that happen to be a subset.

https://github.com/user-attachments/assets/b01daeb3-1768-4d7c-86c0-c6a18168b011

tbar0970:
> Ah okay. But why introduce new terminology? Why not just allow multiple volunteer groups? 

New terminology for a different function.  Say I've got "all 4pm musicians", "4pm band A" and "4pm band B".  For a given date I'd like to be able to quickly assign "4pm band A" (or B).  But for a given date I'd never want to assign "all 4pm musicians".  They are the broader pool.  "Teams" are people to be assigned together.

> But how about doing the 'materializing' in Javascript, visually and up-front?

Hmm, yes, It'd be nice to show the "materializing" early.  But two issues:  People might be in multiple teams,  which would make the tree layout confusing.  And if I choose "team A" but then change my mind, I'd prefer not to have to unselect 5 team members to undo it.  

Perhaps a way to just make it more transparent is to to have the dropdown option list (some) team members:  "Team A [Tom,Jeff,...]"

radarhere:
> It looks like you've still got nested groups, which then show up as teams to be selected?

...I don't think so. I think it should be showing all groups as 'Volunteer Teams' in the interface. Maybe I just need to change the group names in subsequent demo videos to be clearer.

PR #1428, opened 2026-04-11
Add exact person ID matching to getMatchingPerson() for reliable bulk CSV imports

sevasek:
After migrations like Fluro → Jethro, a church may end up with multiple archived duplicate person records. The current CSV import fails with “too many people with the same details” because getMatchingPerson() only does fuzzy matching on name/email/mobile.

This PR adds a tiny, isolated early-exit:
- If 'id' or 'person_id' is supplied in $match_data, it does an immediate exact lookup by the internal person ID (certain match).
- All existing fuzzy logic, scoring, and behaviour for normal imports remains 100% unchanged.

See the added comment in db_objects/person.class.php for details.

Next steps (planned follow-up PRs):
- Expose person.id in reports/CSV exports
- Update the import UI to allow mapping an ID column and pass it through

### DIFF
--- a/db_objects/person.class.php
+++ b/db_objects/person.class.php
@@ -546,11 +546,26 @@ class Person extends DB_Object
 
 	/**
 	 * Find a person who matches the details given.
-	 * @param array $match_data - keys can be first_name, last_name, mobile_tel, email
+	 * @param array $match_data - keys can be first_name, last_name, mobile_tel, email, OR id/person_id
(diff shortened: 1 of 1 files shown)


tbar0970:
I'm OK with this functionality, but be aware that Person::getMatchingPerson() isn't called by the CSV importer.  See view_10_admin__7_import.class.csv _findExistingPerson().

You can already include Person ID in a report.

PR #1431, opened 2026-04-13
Remove 'Show init SQL' button in System configuration page

jefft:
When `JETHRO_VERSION` is undefined or set to `DEV` (i.e. not normally), the System configuration page acquires this 'Show init SQL' button:

I'm wondering what it's for? Is it some artifact of a past debugging session?

If it's useful in any way to any person, great.

### DIFF
--- a/views/view_10_admin__6_system_configuration.class.php
+++ b/views/view_10_admin__6_system_configuration.class.php
@@ -60,17 +60,6 @@ class View_Admin__System_Configuration extends View {
 
 	public function printView()
 	{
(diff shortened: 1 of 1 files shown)


tbar0970:
The button is useful if you've added new db_object classes, which create tables during install, (db_object::getInitSQL) and you're trying to figure out what SQL needs to go in the upgrade script. As you've noticed, it's not shown in production systems.

jefft:
I see, thanks!

Issue #1402, opened 2026-03-13
Reorganize source to separate web mountable files in webroot/


jefft:
Jethro's filesystem struture is very simple:
```
/index.php
/conf.php
/include/
/views/
/resources/
/upgrades/
/vendor/
...
```

Typically you'd just point a webserver like Nginx at this root directory, with a special handler for `*.php`. 

The problem is this exposes files that shouldn't be exposed:

https://easyjethro.com.au/demo/conf.php
https://easyjethro.com.au/demo/scripts/roster_reminder_sample.ini
https://easyjethro.com.au/demo/scripts/migrate_settings.php
https://easyjethro.com.au/demo/upgrades/2024-upgrade-to-2.36-after.php
https://easyjethro.com.au/demo/vendor/composer/LICENSE
https://easyjethro.com.au/demo/vendor/composer/installed.json
https://easyjethro.com.au/demo/vendor/composer/installed.php

I.e.:
 - stuff in `scripts/` and `upgrades/` only intended for CLI use, including `scripts/*.ini` files with sensitive info
 - `conf.php.bak` or `.conf.php.swp` if you or your editor accidentally created that, exposing your MySQL credentials.
 - every file in every vendor dependency, most of which (besides `dist/` subdirs) isn't meant to be randomly executable by end users.

Jethro basically has no dependencies, so it hasn't mattered much, but to evolve further this needs to change. E.g.:
 - I'd like to add some phpunit tests, but can't without bloating the Jethro zip with vendor/phpunit/* files and exposing `tests/*.php` URLs in every Jethro instance.
 - I also have a WIP PR that adds php-cs-fixer, blocked for the same reason.
 - Other features, like SSO support (https://github.com/tbar0970/jethro-pmm/issues/805) require dependencies.

We should rearrange Jethro to be more like other modern PHP codebases, with a distinct subdirectory to mount in a webserver:
```
webroot/index.php
webroot/resources/
src/conf.php
src/include/
src/views/
src/upgrades/
/vendor/
```

Then e.g. webroot/index.php references `../src/conf.php` or `../src/include/general.php`.

(Many projects use `public/` as the name for this webserver-mountable directory. `public/` already means something in Jethro, so I went with `webroot/` which is more descriptive anyway)

[...]

PR #1440, opened 2026-04-26
Regression tests for Jethro

jefft:
Some background: I'm working on a large PR to add PHP 8 type declarations to Jethro, so that IDEs can help more with autocomplete and flagging common errors. It's a risky PR because it touches a lot of code. PHP is a weakly typed language, and although static analyzers like PHPStan can catch a lot, there's still a chance that a typecast or new type declaration will break Jethro.

So alongside my risky PR, I've developed this regression testing framework to validate everything still works. Run it before a risky code change, and after, and it will check if anything changed in Jethro's HTML output, over a representative set of pages.

The regression test is written to test against the [demo data](https://github.com/tbar0970/jethro-pmm/pull/1398) available online and for local download, but is simple enough to modify to test other Jethro instances. Here is a demo, first testing https://easyjethro.com.au/demo/, then running the same data locally, making a local code change, and showing how the regression tests detects the breakage:

[asciicast](https://asciinema.org/a/977436)

In terms of technology, I implemented the tests in https://www.nushell.sh because 
1) nushell is a static binary with no dependencies
2) I don't want to introduce any composer PHP dependencies until #1402 is fixed
3) it's fun and something I'm learning

When #1402 is addressed we can redo functional testing properly in PHP.

I've raised this PR in case other developers find regression testing useful too, but given that it's not PHP, I'm happy for it to be an unofficial thing living in my fork. This PR is mainly a FYI that it exists.


tbar0970:
Automated tests = a worthy project!

If you think a different approach will be better long term, then let's keep this work sitting in this PR for now. Good to have a record of it.

A comment on type declarations: PHP code of Jethro's vintage will, at various points, take advantage of PHP's weak typing. It will be necessary to look carefully for situations where this is the case. An obvious example is functions which return FALSE on error.  I suspect there are more subtle examples in there.

PR #1443, opened 2026-04-29
Move db_object.class.php from include/ to db_objects/, where it more naturally belongs

jefft:
It's always been slightly odd that `db_object.class.php` is in `include/`, while all its children are in `db_objects/`:

```
include/db_object.class.php
db_objects/person.class.php
db_objects/family.class.php
.... 40 child classes ...
```
This PR moves it to its (IMO) logical place. As a side-effect it makes editors and code tools like phpstan happy, as they can now follow codepaths without `include_once` statements. The `include_once` statements can be totally removed.


tbar0970:
There is a pattern in the existing arrangement: All the abstract parent classes db_object, view, call etc are in `includes/`, and their subclasses are in dedicated folders. The theoretical logic is that a hypothetical programmer might make another project using the framwork in the`include` directory but with a whole different set of db_objects, views etc. (Yes, I know nobody is going to do that now).

Happy to remove unnecessary `include` statements.  The include_path set near the top of init.php should make them irrelevant. 
```
set_include_path(ini_get('include_path').$path_sep.JETHRO_ROOT.$path_sep.JETHRO_ROOT.'/include/'.$path_sep.JETHRO_ROOT.'/db_objects/');
```

I'd prefer to leave it as is, but if there's a compelling reason to move db_object.class.php, the same should be applied to view.class.php, call.class.php and possibly some others.

sevasek:

PR #1444, opened 2026-04-29
Consistently capitalize the names of database objects, e.g. 'family' -> 'Family'

jefft:
**This PR builds upon https://github.com/tbar0970/jethro-pmm/pull/1443 which should be dealt with first. Only commit 7af8ff519956db0bc3a136bf3e909a6476b38060 is actually part of this PR**

PHP is case-insensitive but have consistent capitalization helps understanding when reading the code.

This fixes the 'class.nameCase' PHPStan errors.


sevasek:

PR #1445, opened 2026-04-29
Introduce Person::getFamily() helper method

jefft:
**This PR builds upon https://github.com/tbar0970/jethro-pmm/pull/1444, which builds on https://github.com/tbar0970/jethro-pmm/pull/1443, which should be dealt with first. Only the last commit is actually part of this PR**

Introduce Person::getFamily() - new method is used in 8 call sites.

This simplifies code and gives us a nice Family object to work with, rather than the array getDBObject() returns.


sevasek:

PR #1447, opened 2026-04-30
Fix crash caused by trying to save half-populated notes (#1446)

### LINKED ISSUE #1446: Bulk-editing assignee notes breaks with 'History field is not an array'
Going to Notes -> For Immediate Action -> 'Edit the assignees for all these notes', and then clicking 'Save Assignees' causes Jethro to break:

```
Fatal Error (Exception)
History field is not an array - this should not be. Aborting.

Show Details
Line 366 of File /home/jethro/code/current/include/db_object.class.php
	Array
(
    [0] => Array
        (
            [file] => /home/jethro/code/current/db_objects/abstract_note.class.php
            [line] => 295
            [function] => save
            [class] => db_object
            [type] => ->
        )

    [1] => Array
        (
            [file] => /home/jethro/code/current/views/abstract_view_notes_list.class.php
            [line] => 46
            [function] => save
            [class] => Abstract_Note
            [type] => ->
        )

    [2] => Array
        (
            [file] => /home/jethro/code/current/include/system_controller.class.php
            [line] => 125
            [function] => processView
            [class] => Abstract_View_Notes_List
            [type] => ->
        )

    [3] => Array
        (
            [file] => /home/jethro/code/current/index.php
            [line] => 66
            [function] => run
            [class] => System_Controller
            [type] => ->
        )

)
```

https://github.com/user-attachments/assets/de4c82a2-a2aa-405a-9db3-11e91f79efa3

jefft:
Fixes #1446

The root cause of this, and https://github.com/tbar0970/jethro-pmm/issues/836 earlier, is that it is not safe to `populate()` then `save()` an object, as in this code pattern:

```php
// Load a bunch of rows from db into $things
$dummy = new Something();
foreach ($this->things as $id -> $arr) {
  $dummy->reset();
  $dummy->populate($id, $arr);    // loads a subset of fields
  ... // update somehow
  $dummy->save();    // history is lost!
}
```
`populate()` is for quick read-only objects, and does not populate serialised fields like `history`. If you `populate()` an object like an Abstract_Note, then `save()` it, the history gets lost.

We actually had this data loss bug in Jethro (losing note histories) until 2.33, when #836 was found and fixed. #836 is a different instance of `populate()` being misused, but as part of the fix the `"History field is not an array - this should not be. Aborting."` sanity check you see triggered here. Hurrah for sanity checks!

This PR has two commits: one fixes the immediate problem, the other makes `populate()` into less of a footgun by documenting its intended purpose, and adding a flag so that if it ever misused, and `save()` called, it blows up with a clean error.

### DIFF
--- a/include/db_object.class.php
+++ b/include/db_object.class.php
@@ -10,6 +10,7 @@ class db_object
 	protected $_old_values = Array();
 	protected $_held_locks = Array();
 	protected $_acquirable_locks = Array();
+	protected $_populated_partial = FALSE;
 
 	protected $_tmp = Array();
 
@@ -201,6 +202,7 @@ class db_object
 
 		$parent_class =  strtolower(get_parent_class($this));
 		if ($parent_class != 'db_object') {
+            /** @var DB_Object $parent_obj */
 			$parent_obj = new $parent_class();
 			$parent_obj->populate(0, $this->values);
 			if (!$parent_obj->create()) {
@@ -314,9 +316,9 @@ class db_object
 				} else {
 					$this->values[$name] = unserialize($this->values[$name]);
 				}
-					
 			}
 		}
+		$this->_populated_partial = FALSE;
 	}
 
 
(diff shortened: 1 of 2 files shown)


include/db_object.class.php
```diff
@@ -201,6 +202,7 @@ public function create()
 
 		$parent_class =  strtolower(get_parent_class($this));
 		if ($parent_class != 'db_object') {
+            /** @var DB_Object $parent_obj */
```
radarhere:
```suggestion
            /* @var DB_Object $parent_obj */
```

PR #1451, opened 2026-05-01
UI cleanup: View Family Members: add missing action divider

jefft:
On the Family page's "Family Members" section, there is no horizontal line separating the members from the actions:

image

This PR changes it to:

image

This makes it consistent with other places in Jethro:

image

image

### DIFF
--- a/resources/less/jethro.less.php
+++ b/resources/less/jethro.less.php
@@ -1212,6 +1212,8 @@ img.person-photo {
 	row-gap: 10px;
 	box-sizing: border-box;
 	margin-bottom: 10px;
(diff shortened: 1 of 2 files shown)


resources/less/jethro.less.php
```diff
@@ -1212,6 +1212,8 @@
 	row-gap: 10px;
 	box-sizing: border-box;
 	margin-bottom: 10px;
+	padding-bottom: 8px;
+	border-bottom: 1px solid #dddddd;
```
radarhere, line 1216:
```suggestion
	border-bottom: 1px solid #ddd;
```

PR #1455, opened 2026-05-04
Don't break if upgrades/2026-upgrade-to-2.38.sql is run twice

jefft:
Upgrade SQL should be idempotent.


tbar0970:
Happy for this change. But I'd say ensuring every bit of upgrade SQL is idempotent is going to be difficult.

tbar0970:
A worthy goal where possible though.

PR #1456, opened 2026-05-05
Bugfix #1454: enabled menus now immediately reflect the features enabled

### LINKED ISSUE #1454: Updating 'features' does not take effect until next login
In Admin -> System Configuration there is an **Enabled features** section. Checking/unchecking these features causes parts of Jethro to appear or disappear.

However due to caching, the change does not take effect until one logs out and logs back in again. This is pretty confusing at first.

https://github.com/user-attachments/assets/3879d1cb-0e38-4a01-88c3-8bf2e3194769

jefft:
**Note this PR depends on https://github.com/tbar0970/jethro-pmm/pull/1448 which should be dealt with first**

Fixes #1454

This is done by storing a hash of all "dependencies" of our loaded views (i.e. the set of enabled features, and our current permissions), and recalculating the views if our dependencies changed.

### DIFF
--- a/include/system_controller.class.php
+++ b/include/system_controller.class.php
@@ -36,7 +36,7 @@ class System_Controller
 		$path_sep = defined('PATH_SEPARATOR') ? PATH_SEPARATOR : ((FALSE === strpos($_ENV['OS'], 'Win')) ? ';' : ':');
 		ini_set('include_path', ini_get('include_path').$path_sep.$this->_base_dir);
 
-		if (!isset($_SESSION['views'][$base_dir]) || isset($_REQUEST['regen'])) {
+		if (!isset($_SESSION['views'][$base_dir]) || $this->viewsAreStale($base_dir) || isset($_REQUEST['regen'])) {
 			$scanned_views = Array();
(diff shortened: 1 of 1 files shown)


tbar0970:
This looks to me like it will run an extra query to fetch enabled features, on every page view? That seems expensive and unnecessary. The ENABLED_FEATURES setting will already be exported as a constant (see Config_manager::init()). Couldn't we just store ORIGINAL_ENABLED_FEATURES in the session, check if the ENABLED_FEATURES constant is different, and use that to trigger a refresh?

jefft:
> This looks to me like it will run an extra query to fetch enabled features, on every page view? That seems expensive and unnecessary. The ENABLED_FEATURES setting will already be exported as a constant (see Config_manager::init()). Couldn't we just store ORIGINAL_ENABLED_FEATURES in the session, check if the ENABLED_FEATURES constant is different, and use that to trigger a refresh?

Good point! I forgot `ENABLED_FEATURES` becomes a constant. I've revised the commit.

The general pattern of "store a hash of X's dependencies and invalidate X if they change" seems sound, and we hash user permissions as well as `ENABLED_FEATURES`, so a session variable called `views_deps_hash` seems appropriate.

PR #1459, opened 2026-05-17
Automatic fetching of Bible Reading texts in service handouts, via optional https://api.bible integration

jefft:
As background: in Jethro it is common to define a "bible reading" service component, with tokenized 'Handout Title
Format', e.g.:

```
Bible Readings: %SERVICE_BIBLE_READ_ALL%
Bible Reading 1: %SERVICE_BIBLE_READ_1%
Bible Reading 2: %SERVICE_BIBLE_READ_2%
...
```

In the handout this expands to, e.g. 'Bible Reading 1: John 3:16'.

Now, "bible readings" service components can also have their 'Content' set to a token:

```
%SERVICE_BIBLE_READ_ALL_CONTENT%
%SERVICE_BIBLE_READ_1_CONTENT%
%SERVICE_BIBLE_READ_2_CONTENT%
...
```

These expand to the text of the passage, e.g.:

image

For this to work, one must:
1) sign up to https://api.bible, and obtain an API key
2) In Jethro's System configuration page:

 - enable the 'Bible Reading Downloads' feature
     image

 - in the 'Bible Reading Lookup' section, set the 'Bible Api Apikey' to the API key
     image

3) Edit one's "Bible Reading" service components and:
 - Change 'Handout Visibility' from 'Include Title Only' to 'Include Title and Content'
 - set the 'Content' field to the relevant `%*_CONTENT%` token.
      image

4) Edit relevant service runsheets, edit the Bible Reading item, 'Edit item detail' and change 'Show in handout?' to
   'Title and Content'.

For rendering, API.Bible provide a very nice CSS stylesheet, which is used via hyperlink in the 'Printable' view (if the
feature is enabled). We cannot bundle it as copyright is unclear:
image
For non-printable, simpler hand-rolled CSS fitting our l&f is used.

If 'Bible Reading Downloads' feature is turned off or the API key is unset, %*_CONTENT% tokens expand to blank.

There is also a 'Bible Translation Preferred' config setting, populated (once API key is set) from those available from API.Bible. Note that API.Bible has NIV, CSB, KJV, NKJV and many other translations, but lacks ESV, NLT, RSV and NASB.

For testing purposes, a scripts/biblelookup.php script is provided, which allows command-line lookup of bible readings
for different translations, and a list of bibles.

### DIFF
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ System requirements are:
     * with [zip extension](https://www.php.net/manual/en/book.zip.php) enabled
     * with [xmlwriter extension](https://www.php.net/manual/en/book.xmlwriter.php) enabled, to create DOCX files
     * [GD library](https://www.php.net/manual/en/book.image.php) recommended, to manage the size of uploaded photos
-    * with [curl extension](https://www.php.net/manual/en/book.curl.php) enabled, if you intend to use the Mailchimp integration
+    * with [curl extension](https://www.php.net/manual/en/book.curl.php) enabled, if you intend to use the Mailchimp or https://api.bible integration
     * with [exif extension](https://www.php.net/manual/en/book.exif.php) enabled, if you would like to automatically rotate images
 
 The steps to install are:
--- a/calls/call_service_comp_help_runsheet_format.class.php
+++ b/calls/call_service_comp_help_runsheet_format.class.php
@@ -124,6 +124,11 @@ class call_service_comp_help_runsheet_format extends Call
 							<code>%SERVICE_BIBLE_PREACH_ALL%</code> expands to
 							<code>Psalm 111</code>
 						</p>
+						<p>
+							Add <code>_CONTENT</code> to fetch the full passage text instead of just the reference:
+							<code>%SERVICE_BIBLE_READ_1_CONTENT%</code> — fetches and displays the passage
+							using the preferred Bible translation (set in System Configuration).
+						</p>
 					</td>
 				</tr>
 				</tbody>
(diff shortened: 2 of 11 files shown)


tbar0970:
Nice.

> In Jethro's System configuration page:
> enable the 'Bible Reading Downloads' feature

Are you sure we need to define a new "feature"? Entering valid api.bible details could be the trigger that enables this (like with mailchimp)

jefft:
> Are you sure we need to define a new "feature"? Entering valid api.bible details could be the trigger that enables this (like with mailchimp)

I think the logic was: in environments like EJ you might want the URL and API key set in conf.php for security reasons:

image

but you still want to grant administrators the ability to turn the feature on or off.

It is true that this adds a "it's not working because you forgot to turn it on" failure mode, but with status panels (https://github.com/tbar0970/jethro-pmm/pull/1466) the step is more obvious:
image

tbar0970:
> I think the logic was: in environments like EJ you might want the URL and API key set in conf.php for security reasons:
> but you still want to grant administrators the ability to turn the feature on or off.

hmm, yes, I agree you'd want the details saved in conf.php
But if it's configured and working, for what reason would an administrator want to turn it off?
Wouldn't they just not use that keyword in the template doc?
All the other "features" are quite significant things, but this is a deep-down detail.

jefft:
> Wouldn't they just not use that keyword in the template doc?
> All the other "features" are quite significant things, but this is a deep-down detail.

Yes, you're probably right. This being an "external service integration" feature makes it somehow special in my mind, but I can't think why.

I have removed the feature flag and now the feature just works if the url and secret are defined:

image

(showing status panel - https://github.com/tbar0970/jethro-pmm/pull/1466)

Demo online at:

https://push-apibible-6eaf5618.shellbox.dev/?view=admin__system_configuration#bible-reading-lookup
https://push-apibible-6eaf5618.shellbox.dev/?date=2025-08-03&congregationid=2&view=services

PR #1461, opened 2026-05-19
Add sticky sidebar nav to system configuration page

jefft:
The config page is quite long and will get longer (if I have my way :). 

This PR adds a floating menu to the left, linking to setting headings. It is constructed in Javascript at the end of the page render.

The menu vanishes when the viewport gets too small, as on phones.

[jethro_config_nav.webm](https://github.com/user-attachments/assets/a6bea303-70c8-418a-90bf-980670a2d9ad)


tbar0970:
Nice. Id been thinking we needed something like this.

PR #1463, opened 2026-05-21
Use a pencil icon to represent a note object

jefft:
In the Notes tab, Jethro displays a person (or family) icon alongside each note:

image

I think a pencil icon would make more sense:

image

since that is what is used to indicate a note elsewhere:

image

image

### DIFF
--- a/templates/single_note.template.php
+++ b/templates/single_note.template.php
@@ -38,7 +38,7 @@ include_once 'urllinker.php';
 		<?php
 	}
 	?>
-	<i class="icon-<?php echo $type == 'family' ? 'home' : 'user'; ?>"></i>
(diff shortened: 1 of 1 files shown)


tbar0970:
Using the pencil icon for everything would make it impossible to tell the difference between a family note and a person note. This would lead to confusion ("why do these two people have identical notes?"). If there was some kind of person-pencil hybrid, and house-pencil hybrid, I'd be up for that.

jefft:
> Using the pencil icon for everything would make it impossible to tell the difference between a family note and a person note.

I see - family notes appear on the Person notes tab too! I hadn't noticed that. Thanks.

PR #1466, opened 2026-05-23
System configuration page: add section help/status panels

jefft:
In Jethro, external integrations (SMS, SMTP, Mailchimp, etc) are configured in sections of the System configuration page. However there is no feedback as to whether the settings are correct.

This change allows PHP snippets to be defined for any section (sms, smtp, mailchimp, bible_api, etc), showing information and the 'status' of the settings (e.g. the SMS balance), as well as configuration-adjacent one-off operations, like (for SMS) registering a Sender ID.

Currently implemented:
 - 'SMS Gateway' - prints '✓ Gateway is configured.' if correctly configured 
        image

 - 'SMTP Email Server' - prints '✓ SMTP server is reachable and responding. (details)'
        image
 - 'Mailchimp Sync' - prints '✓ Connected to Mailchimp. 3 audience(s) found. (details)'
        image

These 'status panels' load asynchronously (AJAX), and are implemented as Calls:

```php
calls/call_admin_statuspanel_sms.class.php
calls/call_admin_statuspanel_smtp.class.php
calls/call_admin_statuspanel_mailchimp.class.php
```

Being asynchronous, they won't slow the page load if an external system is down, and being implemented as separate calls lets us add new help sections as settings are added, without hardcoding anything in View_Admin__System_Configuration. For instance, if [API.Bible integration](https://github.com/tbar0970/jethro-pmm/pull/1459) is added:

The implementation in View_Admin__System_Configuration is as follows: for each setting with a heading encountered (say, SMS_MAX_LENGTH), the first part of the symbol is isolated ('SMS') and is used to search for a matching Call class ('call_admin_statuspanel_sms.class.php'). The longest prefix wins, e.g. 'BIBLE_API_URL' would use call_admin_statuspanel_bible_api.class.php before call_admin_statuspanel_bible.class.php

The status_panel Calls all extend an abstract Call subclass, Call_Admin_Statuspanel that implements run(), enforcing a common HTML layout. Each status panel Call must implement:

```php
getHelpText()    // Help text displayed at the top of the panel
isConfigured()   // Whether the config settings are present
getStatus()      // Live operational check — is the feature actually working right now?
getOperations()     // Operations the admin can initiate from this status panel.
```

### DIFF
--- a/calls/call_admin_statuspanel.class.php
+++ b/calls/call_admin_statuspanel.class.php
@@ -1,0 +1,277 @@
+<?php
+
+require_once __DIR__ . '/../include/general.php';
+
+/**
+ * Abstract base for admin status panel calls.
+ *
+ * An admin 'status panel' is a HTML snippet reporting the status of some configurable Jethro subsystem, such as SMS, SMTP or Mailchimp. The status panels rendered by subclasses are displayed on the System Configuration page just below the heading ('SMS Gateway', etc), invoked via AJAX.
+ *
+ * The status panel consists of:
+ *
+ *  - a 'Status:' summary line
+ *  - Enabled/Disabled
+ *  - Configured
+ *  - Status message
+ *  - Operations
+ *
+ * Where:
+ *
+ * "Status: …" is a plain-English summary ({@see self::summaryText()}), e.g.
+ * "Status: enabled but not configured".
+ *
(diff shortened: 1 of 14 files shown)


jefft:
I think this PR is good for review.

PR #1470, opened 2026-05-25
Notes: add checkboxes allowing filtering by status/assignee

jefft:
Persons sometimes have hundreds of Notes, making it hard to find relevant ones. This commit allows Notes to be filtered, by clicking a set of checkboxes on the right:

```markdown
[ ] No Action Required
[ ] Requires Action
[ ] Failed
[ ] Complete
----
[ ] Assigned to me
```

Visually this is done in a sidebar, and the 'Add Note' button comes along for the ride.

image

Filtering is implemented for both Person and Family notes (reusing code).

[jethro_notes_filtering.webm](https://github.com/user-attachments/assets/fbc837a6-0a63-41b2-b988-d116114920d9)

### DIFF
--- a/templates/note_filters.template.php
+++ b/templates/note_filters.template.php
@@ -1,0 +1,28 @@
+<?php
+/**
+ * Note sidebar: filter checkboxes and action links for the Notes section.
+ *
+ * Expected variables:
+ * @var string|null $add_note_html  HTML for the "Add Note" link, or null
+ */
+$GLOBALS['system']->includeDBClass('abstract_note');
+$statusDummy = new Abstract_Note();
+$statusOptions = $statusDummy->fields['status']['options'];
+?>
+<div class="panel-sidebar pull-right">
+	<i><?php if (!empty($add_note_html)) echo "<i>".$add_note_html."</i>"; ?></i>
+	<fieldset class="hidden-phone">
+		<legend><?php echo _('Status:'); ?></legend>
+		<?php foreach ($statusOptions as $value => $label): ?>
+		<label class="checkbox">
+			<i><input type="checkbox" class="note-status-filter" value="<?php echo $value; ?>" checked>
+			<?php echo ents($label); ?></i>
+		</label>
+		<?php endforeach; ?>
+		<legend><?php echo _('Assignee:'); ?></legend>
+		<label class="checkbox">
+			<i><input type="checkbox" id="note-assignee-filter">
+			<?php echo _('Assigned to me'); ?></i>
+		</label>
+	</fieldset>
+</div>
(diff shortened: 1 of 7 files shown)


templates/note_filters.template.php
```diff
@@ -0,0 +1,53 @@
...
+}
+.note-sidebar legend {
+	display: inline;
+	width: auto;
+    margin: 1em 0 0.5em 0;
+    padding-right: 8px;
+	font-size: 0.85em;
+    /*border-bottom: 0px;*/
+    /*border-top: 1px;*/
+
```
radarhere:
```suggestion
```

templates/note_filters.template.php
```diff
@@ -0,0 +1,53 @@
...
+}
+</style>
+<div class="note-sidebar pull-right">
+	<fieldset>
+        <?php if (!empty($add_note_html)): ?>
+            <legend></legend> <?php // Don't delete ?>
```
radarhere: Would it be more helpful to have the comment explain why it shouldn't be deleted? I expect it is for visual alignment, but if that were completely obvious, then my opinion would be that the comment isn't needed in the first place.

templates/note_filters.template.php
jefft: Fixed, thanks

templates/note_filters.template.php
jefft: It was to prevent 'Add Note' from mysteriously jumping below the 'Status' legend - but you're right, ugly. I've eliminated it in the latest push. Thanks!

tbar0970:
Can we please move the JS to jethro.js and the CSS to jethro.less.php

And: How does this behave on small (phone) screens? Simplest option might be to hide it.

tbar0970:
(I’m up for this feature, but I haven't seen persons with 100s of notes... those pastors must have been super busy!)

jefft:
> (I’m up for this feature, but I haven't seen persons with 100s of notes... those pastors must have been super busy!)

Churches that heavily use action plans, and who send SMSes, have lots:

image

Excluding 'SMS Sent' I've got 47 notes (collected over 13 years): prayer points, church directory permissions, informal surveys, etc.

jefft:
> Can we please move the JS to jethro.js and the CSS to jethro.less.php

Sure, done

And: How does this behave on small (phone) screens? Simplest option might be to hide it.

I didn't think of that.. 

The sidebar now hides, except for the 'Add Note' button, which I figure is important.

[jethro_note_sidebar_small.webm](https://github.com/user-attachments/assets/66b5466a-9eb8-484f-a8cb-cd90e163a016)

jefft:
I have adjusted the CSS to make the sidebar sticky (always visible). Often you'll be scrolled halfway down a long page before you notice the 'Complete' notes are swamping what you really want to see.

[jethro_note_sidebar_sticky.webm](https://github.com/user-attachments/assets/299b31c0-46f1-4e89-a993-18350db75093)

PR #1489, opened 2026-07-20
Add optional Reply-To configuration to reminder scripts

jefft:
Jethro hosters like easyjethro.com.au send emails on behalf of real churches. For example, they might send a 'WWCC Renewal Reminder' to church members:

  From: Jethro <noreply@easyjethro.com.au> 
  Subject: WWCC Renewal Reminder

Replies ought to go to a church Safe Ministry Rep. Previously this was only possible by faking the From: address:

```email
  From: Safe Ministry <safeministry@example.church> 
  Subject: WWCC Renewal Reminder
```

however in recent years almost all mail clients will reject this as spam unless DKIM + SPF records are set up to allow the Jethro hoster to send 'from' example.church.

This PR solves the problem by allowing a Reply-To: header to be set:

```email
  From: Jethro <noreply@easyjethro.com.au> 
  Subject: WWCC Renewal Reminder
  Reply-to: Safe Ministry <safeministry@example.church>
```

This is implemented for all email-sending scripts:

- date_reminder.php: REPLY_TO_ADDRESS and REPLY_TO_NAME from the INI file. Applied to both the individual reminder and the supervisor summary email paths.

- roster_reminder.php: REPLY_TO_ADDRESS and REPLY_TO_NAME from the INI file. Applied in all four email-sending sites (Emailer path for main email + summary; raw mail() path for main email + summary).

- task_reminder.php: TASK_NOTIFICATION_REPLY_TO_ADDRESS and TASK_NOTIFICATION_REPLY_TO_NAME from conf.php. Applied to the per-assignee notification loop.

All settings are optional — when absent, no Reply-To header is omitted, so existing configurations work unchanged.

Sample INI files (date_reminder_sample.ini, roster_reminder_sample.ini) include commented-out examples.

### DIFF
--- a/scripts/date_reminder.php
+++ b/scripts/date_reminder.php
@@ -84,6 +84,9 @@ if (!empty($ini['SMS_MESSAGE'])) {
 	}
 }
 require_once JETHRO_ROOT.'/include/emailer.class.php';
+$replyToAddress = !empty($ini['REPLY_TO_ADDRESS']) ? $ini['REPLY_TO_ADDRESS'] : null;
+$replyToName    = !empty($ini['REPLY_TO_NAME'])    ? $ini['REPLY_TO_NAME']    : null;
+
 
 // Send individual reminders and collate summary info
 $summaries = Array();
@@ -119,6 +122,10 @@ foreach ($summaries as $supervisors => $remindees) {
 	} else {
 		$message->setTo(explode(';', $supervisors));
 	}
+
+	if ($replyToAddress) {
+		$message->setReplyTo($replyToAddress, $replyToName);
+	}
 	$res = Emailer::send($message);
 	if (!$res) {
 		echo "Failed to send summary email to $supervisors \n";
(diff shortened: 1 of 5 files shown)


tbar0970:
I agree that reply-to is the right approach
But didn't we already implement this via https://github.com/tbar0970/jethro-pmm/commit/3057e97e36e09da3f06475d2c66dbff39197ac36 ??

jefft:
> But didn't we already implement this via https://github.com/tbar0970/jethro-pmm/commit/3057e97e36e09da3f06475d2c66dbff39197ac36 ??

Why yes! I'd not seen `OVERRIDE_EMAIL_FROM` before, but it achieves the same thing, in a nicer way. I can set just once in `conf.php`:
```conf
define('OVERRIDE_EMAIL_FROM', 'noreply@easyjethro.com.au');
```
and then each of the many of `date_reminder` ini scripts can set `FROM_ADDRESS` how they like:

```conf
FROM_ADDRESS = "safeministry@church.org"
```
and no matter what the ini file sets the result is always DKIM-safe:
```mail
From: noreply@easyjethro.com.au
Reply-To: safeministry@church.org
```

I'll close this PR and open a new one addressing a few niggles. Thanks!

jefft:
Succeeded by https://github.com/tbar0970/jethro-pmm/pull/1490

PR #1491, opened 2026-07-21
date_reminder: allow filtering by person status

jefft:
The date_reminder.php script is useful for sending reminders of things like upcoming WWCC expiries. But sometimes we'd only want to email 'core' people. This PR adds a PERSON_STATUS field that limits who is notified to people with certain statuses.

If a non-existent status is specified, the script prints a warning but still continues. Non-integer text is ignored.

### DIFF
--- a/scripts/date_reminder.php
+++ b/scripts/date_reminder.php
@@ -6,6 +6,9 @@
  * It can also send a summary of the reminders sent, to people in the same congregation
  * with a certain person status.
  *
+ * An optional PERSON_STATUS setting (comma-separated person_status IDs) limits
+ * recipients to persons with one of those statuses.
+ *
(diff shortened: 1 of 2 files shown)


scripts/date_reminder.php
```diff
@@ -40,6 +43,22 @@
 $GLOBALS['system'] = System_Controller::get();
 //error_reporting(E_ALL);
 
+$statusFilter = [];
+if (!empty($ini['PERSON_STATUS'])) {
+	$allStatusIds = array_column($GLOBALS['db']->queryAll('SELECT id FROM person_status'), 'id');
```
tbar0970, line 48: JethroDB::queryCol would be cleaner here. For future reference.

PR #1496, opened 2026-07-23
Allow Jethro to be symlinked into an alternate/safer webroot

jefft:
Jethro's source distribution does not have a distinct web server document root (often called public/, web/, /webroot, /htdocs etc).  Simply deploying the source distribution exposes /vendor/, /scripts/ and other unwanted content. This problem is documented in #1402

This PR allows for a workaround to #1402, and an overall nice deployment approach: let users create their own webroot, e.g. `/srv/www/html/jethro/` containing symlinks to Jethro's source distribution files, but only for files/directories that ought to be in a webroot (not `upgrades/`, `scripts/` etc). For instance, we might have:

```sh
/srv/www/html/jethro/      # webroot that nginx/apache reads
/usr/src/jethro/2.38.0/    # Jethro source
/usr/src/jethro/current/   # symlink to ../2.38.0

/srv/www/html/jethro/index.php   # symlink to ../current/index.php
/srv/www/html/jethro/include/    # symlink to ../current/include/
...
/srv/www/html/jethro/conf.php    # NOT a symlink - our live config
```
To upgrade Jethro, just unzip the new version and update /usr/src/jethro/current/ symlink. Since conf.php lives only in our webroot it persists across upgrades.

This also makes separate test deployments possible. E.g. one can use https://www.jetify.com/devbox nginx + php-fpm + mariadb for development locally: create symlinks from `devbox.d/web/` to `../../`, and have a devbox.d/web/conf.php pointing to the Devbox mariadb instance.

So symlink deployments are great, but to work, Jethro can't use the real location (`__FILE__` / `__DIR__`) of index.php as its `JETHRO_ROOT` (used to find conf.php initially) - it must use the symlink location. This PR achieves that by using the `SCRIPT_FILENAME` superglobal ("the absolute pathname of the currently executing script") instead of `__DIR__` to figure out `JETHRO_ROOT`.  `SCRIPT_FILENAME` is set in CLI and web contexts. It should work just the same for the non-symlink case.

### DIFF
--- a/index.php
+++ b/index.php
@@ -24,7 +24,7 @@
  * @package jethro-pmm
  */
 
-define('JETHRO_ROOT', dirname(__FILE__));
+define('JETHRO_ROOT', dirname($_SERVER['SCRIPT_FILENAME']));
 define('TEMPLATE_DIR', JETHRO_ROOT.'/templates/');
 
 // Load configuration
--- a/members/index.php
+++ b/members/index.php
@@ -24,13 +24,13 @@
  * @package jethro-pmm
  */
 
-define('THIS_DIR', str_replace('\\', '/', dirname(__FILE__)));
+define('THIS_DIR', str_replace('\\', '/', dirname($_SERVER['SCRIPT_FILENAME'])));
 define('JETHRO_ROOT', preg_replace('#/members$#', '', THIS_DIR));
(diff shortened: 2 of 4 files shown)


tbar0970:
NB in the list of symlinks in your top post, I think "include" should actually be "resources". This is where JS, CSS etc lives.

Issue #1088, opened 2024-10-24
Jethro sessions timing out too early


jefft:
Jethro has a 'Session Timeout Mins' system configuration option that claims to determine how long a user can be inactive for, before they need to log in again:

![image](https://github.com/user-attachments/assets/8a0f787d-199d-41b7-a902-119d24872f18)

It isn't working though. Sessions expire after about 1h of inactivity. This can be seen by logging the `X-JethroSession` response header (added in https://github.com/tbar0970/jethro-pmm/pull/851):
```
LogFormat "%V:%p %a %{X-Jethro-Session}o %{X-Jethro-User}o %t \"%r\" %>s %O \"%{Referer}i\" \"%{User-Agent}i\" reqtime=%D uid=%{X-Jethro-UID}o" vhost_combined_timed
```

For one particular user, the session changes (indicating a timeout) after about an hour of inactivity:
```
root@server/var/log/apache2/jethro # cat access.log | awk '$4=="tom" { print $3, $5}' 
ue4p0 [24/Oct/2024:09:56:02
ue4p0 [24/Oct/2024:09:56:05
ue4p0 [24/Oct/2024:09:56:09
ue4p0 [24/Oct/2024:09:56:19
         1h 2m break
9ah0o [24/Oct/2024:10:58:24
9ah0o [24/Oct/2024:10:58:26
...9ah0o [24/Oct/2024:11:06:19
         1h 51m break
v3mrs [24/Oct/2024:12:57:40
v3mrs [24/Oct/2024:12:57:40
v3mrs [24/Oct/2024:12:57:40
v3mrs [24/Oct/2024:12:58:55
v3mrs [24/Oct/2024:12:58:57
         44m 26s break
haccr [24/Oct/2024:13:43:23
         33s break     <-- I can't explain this one
pllck [24/Oct/2024:13:43:56
pllck [24/Oct/2024:13:45:07
        2h 7m break 
lq32o [24/Oct/2024:15:52:52
```

jefft:
The problem is that PHP has its own session timeout mechanism based on the [`session.gc_maxlifetime`](https://www.php.net/manual/en/session.configuration.php#ini.session.gc-maxlifetime) configuration option. Jethro is not setting `session.gc_maxlifetime` to reflect `SESSION_TIMEOUT_MINS`.

[...]

Issue #949, opened 2023-09-05
Generate and use thumbnail of person and family images


jefft:
It would be nice if Jethro would resize uploaded Person / Family images to a reasonable size before storing and serving them in queries. 

Currently there is no size limit to uploaded photos, except that which PHP imposes (`upload_max_filesize` and `post_max_size`).  We have an EJ instance where every Family record (and many Person records) have 7Mb images attached.  This causes listing Persons and Families to be slow, as you'd expect, but the really catastrophic thing is generating a Family Contact List, as either HTML or DOCX. This generated a 757Mb response, and generated 6Gb of persistent temp files (see #947), presumably due to aborted attempts:

![image](https://github.com/tbar0970/jethro-pmm/assets/205995/c9e94af0-e5f3-42a0-a341-7c1c6d947fac)

The Jethro database is also 840Mb, 10x more than a typical large instance. 

A workaround would be to lower `upload_max_filesize` or `post_max_size` to restrict image sizes, but users do sometimes need to attach large'ish attachments to Mailchimp sends.

Issue #152, opened 2016-03-02
Person Report fails due to max_input_vars PHP limitation


jefft:
Our Jethro has a large number of custom fields. If I go to Person -> Reports -> 'Create a new report', leave all settings as default (i.e. a temporary report) and click 'Save and view results', I get an error:

   Error: Name is a required field for Person_Query and cannot be left empty

and in the database, a warning followed by errors:

````
PHP Warning:  Unknown: Input variables exceeded 1000. To increase the limit change max_input_vars in php.ini. in Unknown on line 0, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
Undefined index: save_option - Line 570 of /srv/www/coastec.net.au/jethro/2.14.2/db_objects/person_query.class.php, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
Undefined index: show_fields - Line 644 of /srv/www/coastec.net.au/jethro/2.14.2/db_objects/person_query.class.php, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
Invalid argument supplied for foreach() - Line 14 of /srv/www/coastec.net.au/jethro/2.14.2/include/general.php, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
Undefined index: group_by - Line 647 of /srv/www/coastec.net.au/jethro/2.14.2/db_objects/person_query.class.php, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
Undefined index: sort_by - Line 654 of /srv/www/coastec.net.au/jethro/2.14.2/db_objects/person_query.class.php, referer: https://jethro.coastec.net.au/?view=persons__reports&queryid=0&configure=1
````

jefft:
The problem came up before with Issue #6. I'm not sure what a long-term fix is, but for now, I've set `max_input_vars = 10000` in `/etc/php5/apache2/php.ini`.

tbar0970:
The system in question has ~100 custom fields, which is fairly unusual!

PR #1497, opened 2026-07-23
Add Nagios health check and status.php script showing Jethro runtime status

jefft:
Add Nagios health check and status.php script showing Jethro runtime status

This adds a /status.php script, returning JSON with useful info about the Jethro deployment, e.g.:

```json
{
  "user": "jethro",
  "php_compatible_version": true,
  "max_input_vars": 1000,
  "xdebug_loaded": false,
  "xdebug.mode": false,
  "xdebug.start_with_request": false,
  "xdebug_performance_impact": true,
  "opcache_enabled": true,
  "post_max_size": 8,
  "upload_max_filesize": 2,
  "db_charset_results": "utf8mb4",
  "session_gc_maxlifetime": 5400,
  "session_gc_probability": 1,
  "session_gc_misconfigured": true,
  "php_extensions_installed": true,
  "mod_unique_id_loaded": true,
  "log_errors": "0"
}
```

`/status.php` is intended to be queried by a monitoring system like Nagios. A Nagios-compatible
`scripts/check_http_jethrostatus` is also provided that translates the output into healthy/not healthy status.

For the JSON above, scripts/check_http_jethrostatus would report:

```
WARNING - max_input_vars is 1000, must be >= 10000; post_max_size is 8M, must be >= 20M; upload_max_filesize is 2M, must be >= 20M
```

`/status.php` returns lots of details beyond just "it's up", because there are lots of ways to mess up a Jethro
installation, particularly across upgrades. Of note in the JSON:

| Field                       | Expected | Problem                                                      | Explanation                                                    |
|-----------------------------|----------|--------------------------------------------------------------|----------------------------------------------------------------|
| `user`                      | "jethro" | "root"                                                       | Ensure PHP is running as the correct OS user (e.g. not root)   |
| `php_compatible_version`    | true     | "PHP 8.0.30 is not > 8.1"                                    | Ensure the expected PHP version is used                        |
| `php_extensions_installed`  | true     | "Missing extensions: exif, gd"                               | Jethro requires some extensions (`curl`, `gd`, `zip`, etc) as documented in the README.md |
| `max_input_vars`            | 10000    | 1000                                                         | PHP restricts the number of params submitted, defaulting to 1000. This is too few for most Jethros when marking large attendances (#6) or running reports on systems with many custom fields (#152). The max_input_vars should almost always be increased from its default |
| `post_max_size`             | 20       | 8                                                            | PHP `post_max_size` governs the HTTP POST size, with 8Mb being the default. Sometimes contact photos exceed 8Mb (see related #949), so the operator might want to increase `post_max_size` and validate it is increased |
| `upload_max_filesize`       | 20       | 2                                                            | Similar to `post_max_size`, often the default 2MB is too small |
| `db_charset_results`        | `utf8mb4`| `utf8`, `latin1`                                             | Old MySQL defaulted to `utf8` which isn't real UTF-8 and causes a giant mess on upgrades. See [this comment](https://github.com/tbar0970/jethro-pmm/pull/754#issuecomment-2382842676) and [this](https://github.com/tbar0970/jethro-pmm/issues/1088#issuecomment-2436805398). |
| `opcache_enabled`           | true     | "opcache JIT is enabled"                                     | opcache JIT is pointless for apps like Jethro where the PHP process lifetime is one request — it just burns CPU on warmup. Returns `true` when JIT is off (good), error string when on. |
| `session_gc_misconfigured`  | true     | "Session GC misconfigured: gc_probability=0, gc_maxlifetime=1440s < ..." | Detects the Debian/Ubuntu `sessionclean` trap: the system cron job handles GC instead of PHP (`gc_probability=0`), but `gc_maxlifetime` is left at the low default (1440s). Sessions then expire before `SESSION_TIMEOUT_MINS`, logging users out prematurely. Fix: raise `session.gc_maxlifetime` to at least `SESSION_TIMEOUT_MINS*60`. Ref [#1088](https://github.com/tbar0970/jethro-pmm/issues/1088). |
| `mod_unique_id_loaded`      | true     | "mod_unique_id Apache module not enabled"                    | Apache's `mod_unique_id` generates a unique-per-request ID forwarded as `X-Request-ID`. Jethro uses this for log correlation and request tracing. Without it, requests can't be individually distinguished in logs. |
| `xdebug_performance_impact` | true     | "xdebug performance impact: mode=debug, start_with_request=yes" | Xdebug adds significant overhead even when not actively debugging. This flags when xdebug is loaded AND `mode` is not `off`/empty AND `start_with_request` is not `no` — meaning virtually every request pays a performance penalty. Returns `true` when safe, error string when performance is impacted. |

### DIFF
--- a/scripts/check_http_jethrostatus
+++ b/scripts/check_http_jethrostatus
@@ -1,0 +1,223 @@
+#!/usr/bin/env perl
+# check_http_jethrostatus - Nagios-compatible check for Jethro status.php
+# Fetches JSON from a Jethro status.php endpoint and validates PHP settings
+# critical to Jethro operation.
+#
+# Usage:
+#   check_http_jethrostatus -u http://localhost:8081/status.php
+
+use strict;
+use warnings;
+use Getopt::Long;
+use LWP::UserAgent;
+use JSON::PP ();
+use constant {
+    OK       => 0,
+    WARNING  => 1,
+    CRITICAL => 2,
+    UNKNOWN  => 3,
+};
+
+# ── defaults ────────────────────────────────────────────────────────────────
+my $opt_url;
+my $opt_timeout     = 15;
+my $opt_insecure    = 0;
+my $opt_help        = 0;
+my $opt_max_input_vars      = 10000;
+my $opt_post_max_size       = 20;       # MB
+my $opt_upload_max_filesize = 20;       # MB
+my $opt_db_charset_results  = 'utf8mb4';
+my $opt_require_jethro_version = 0;
+my $opt_required_runtime_user = undef;
+
+GetOptions(
+    'url|u=s'                   => \$opt_url,
+    'timeout|t=i'               => \$opt_timeout,
+    'insecure|k'                => \$opt_insecure,
+    'help|h'                    => \$opt_help,
+    'max-input-vars=i'          => \$opt_max_input_vars,
+    'post-max-size=i'           => \$opt_post_max_size,
+    'upload-max-filesize=i'     => \$opt_upload_max_filesize,
+    'db-charset-results=s'      => \$opt_db_charset_results,
+    'require-jethro-version'    => \$opt_require_jethro_version,
+    'expected-runtime-user=s'   => \$opt_required_runtime_user,
+) or exit UNKNOWN;
+
+if ($opt_help || !$opt_url) {
+    print "Usage: $0 -u <url> [-k] [-t <seconds>] [--max-input-vars N]\n";
+    print "       [--post-max-size N] [--upload-max-filesize N]\n";
+    print "       [--db-charset-results CHARSET]\n";
+    print "\n";
+    print "  -u, --url     URL to status.php (required)\n";
+    print "  -k, --insecure  Skip SSL certificate verification\n";
+    print "  -t, --timeout   HTTP request timeout (default: $opt_timeout)\n";
+    print "  --max-input-vars          Minimum max_input_vars (default: $opt_max_input_vars)\n";
+    print "  --post-max-size           Minimum post_max_size in MB (default: $opt_post_max_size)\n";
+    print "  --db-charset-results      Required db_charset_results (default: $opt_db_charset_results)\n";
+    print "  --require-jethro-version  Require jethroversion field in response\n";
+    print "  --expected-runtime-user   Required OS user (e.g. 'jethro', not 'root')\n";
+    exit UNKNOWN;
+}
+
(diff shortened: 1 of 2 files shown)


tbar0970:
I can see the usefulness of this. But are there security pitfalls if it is
publicly accessible? For example, hiding the php version is now standard
server config, but this script exposes it.

Perhaps better if status.php just returns “all good”, “minor concern” or
“serious issue”?

On a related but distinct note: perhaps some of these checks should be done
within Jethro, eg before/after the install process, or on the system config
page, for an admin user to notice.

jefft:
> I can see the usefulness of this. But are there security pitfalls if it is
> publicly accessible? For example, hiding the php version is now standard
> server config, but this script exposes it.
>
> Perhaps better if status.php just returns “all good”, “minor concern” or “serious issue”?

Yes, you're probably right. I've amended the code and the PR description above. The server now does more interpretation. E.g. `"php_compatible_version": true` reflecting whether the runtime is PHP 8.1+. Instead of listing every PHP extension loaded, it just prints `php_extensions_installed: true`, or enumerates what is missing if not.

> On a related but distinct note: perhaps some of these checks should be done within Jethro, eg before/after the install process, or on the system config page, for an admin user to notice.

In practice, `/status.php` has saved me a number of times after server moves or Ubuntu PHP version updates, where I forgot to apply all necessary `php.ini` tweaks to the new system. Once I forgot to turn off xdebug after debugging a hairy problem in production, without realising the performance hit. My ability to screw things up isn't limited to installation time, so it's worthwhile (being cheap) to check the sanity of the PHP environment at the same time as checking general up'ness.

tbar0970:
> I've amended the code and the PR description above. The server now does more interpretation. E.g. "php_compatible_version": true reflecting whether the runtime is PHP 8.1+. Instead of listing every PHP extension loaded, it just prints php_extensions_installed: true, or enumerates what is missing if not.

Nice.

>My ability to screw things up isn't limited to installation time, so it's worthwhile (being cheap) to check the sanity of the PHP environment at the same time as checking general up'ness.

Yep. Nagios monitoring is good. But I was thinking that Joe Bloggs who doesn't use Nagois might benefit from having such warnings in his face at install time and/or when he goes to the system config page.

jefft:
> Yep. Nagios monitoring is good. But I was thinking that Joe Bloggs who doesn't use Nagois might benefit from having such warnings in his face at install time and/or when he goes to the system config page.

Yes, true. The proper solution there is to make a Docker image or something, so most people get a correct environment by default.  Also, docs. I have PRs on those things pending.

Thanks for committing.

PR #1499, opened 2026-07-27
Fix undefined variable $link_class warning in person_list.template.php

jefft:
"Also, there was an unused $link_class variable that complicated the code, so it is now removed.", said someone on https://github.com/tbar0970/jethro-pmm/pull/1469

It turns out the variable was needed. Without it we get `E_WARNING "Undefined variable $link_class"` on the persons__list_all view and anywhere else the template is included.

image

Also fix docblock typo: $link_Class -> $link_class.

### DIFF
--- a/templates/person_list.template.php
+++ b/templates/person_list.template.php
@@ -3,10 +3,11 @@
 @var $persons
 @var $special_fields
 @var $show_actions	Whether to show the action links for each person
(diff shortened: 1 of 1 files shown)


tbar0970:
Ha. I now know that I should question your critiques of my old code!

PR #1500, opened 2026-07-29
Add Devbox config so devs can quickly get a working dev environment

jefft:
Jethro is rather fiddly software to get up and running in a dev environment. You need to install php + extensions, a web server, php-fpm, database + db user, and tie them all together just right, and ~~none of this is documented~~ it is minimally documented in the README.

Having a Docker `docker-compose.yml` would fix part of the problem, but in development I often want to run SQL against the database, and database-using scripts, and I find having a separate docker world, enterable only via `docker exec`, annoying. 

I've found [Devbox](https://www.jetify.com/docs/devbox) to meet this need rather well. It uses Nix infrastructure to construct a development environment containing all the right tools - php, mariadb, nginx etc, configured in a simple `devbox.json` file. Devbox also lets you define what long-running services (mariadb, nginx, php-fpm, etc) you want, in `process-compose.yml`.

This change adds devbox config files I've developed to provide a full development environment, such that running `devbox services up -b` brings up Jethro setup wizard at http://localhost:8081.

Here is a demo using a https://shellbox.dev VM:

[jethro_devbox.webm](https://github.com/user-attachments/assets/8c9887d9-9a48-4343-9242-de2455d4ce84)

Specifically this change adds:

 - a DEVELOPMENT_DEVBOX.md doc explaining how devs can use Devbox to quickly get started.
 - a tweaked version of the devbox install script (https://get.jetify.com/devbox) installing the right edition.
 - a devbox.json declaring minimal dependencies (php + webserver + db)
 - devbox.d/web/ webroot for the devbox nginx to point to
 - process-compose.yml adding hooks to to the startup to create a jethro database and user.


jefft:
I think this is good to go, along with the docs and functional tests depending on it.

sevasek:

PR #1507, opened 2026-07-30
Functional testing infrastructure

jefft:
`./devbox services up` now brings up a Jethro instance especially for functional testing at http://localhost:8089 (port customizable in devbox.json). This Jethro instance is distinct from the demo/dev instance at http://localhost:8081.

Running `./devbox functests` then runs the Playwright functional tests defined in tests/functional/playwright.config.ts. Currently there is just one login test.

The functest Jethro is prepopulated with realistic sample data from tests/functional/testdata/demodata.sql.gz, derived from https://github.com/tbar0970/jethro-pmm/pull/1398. Since realistic sample data is useful for development, not just functional testing, it's possible to load the sample data into the http://localhost:8081/ instance by running `./devbox run demodata`.

There are two actual functional tests implemented:

1. 'lookaround' is a read-only poke through Jethro prepopulated with demo data.
2. 'walkthrough' goes through the setup wizard and creates some of the demo data. It then incorporates the same read tests as 'lookaround'

Currently these tests fail due to #1511, and is fixed (in a follow-up merge commit) by #1517.


sevasek:

PR #1508, opened 2026-07-30
Add scripts to process LESS files into CSS and generate bundled js for releases.

jefft:
The frontend to this is `./devbox run compile`, which invokes:

 - devbox.d/bin/jethro_compile_js
 - devbox.d/bin/jethro_compile_css

LESS processing is done with lessc, a Javascript tool, so devbox has the 'bun' JS runtime added, and devbox.d/bin/jethro_compile_css runs 'bunx lessc' to compile the *.less files into a single jethro-$VER.css

jethro_compile_js merges *.js into one jethro-$VER.js

### DIFF
--- a/devbox.d/bin/jethro_compile_css
+++ b/devbox.d/bin/jethro_compile_css
@@ -1,0 +1,35 @@
+#!/bin/bash -eu
+
+# Generates resources/css/jethro-$version.css by minifying less output
+
+main() {
+  local version=${1:-${JETHRO_VERSION:-}}
+  [[ -n "$version" ]] || { echo >&2 "No version supplied: pass one as an argument or set JETHRO_VERSION"; exit 1; }
+  #devbox init
+  #devbox -q add nodejs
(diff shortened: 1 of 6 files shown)


sevasek:

devbox.d/bin/jethro_compile_css
```diff
@@ -0,0 +1,35 @@
...
+  bun add less@1.7.5
+  rm -rf buildenv
+  mkdir buildenv
+  # A fake conf.php to keep jethro.less.php happy
+  touch buildenv/conf.php
+  cp -r resources/ buildenv  # a symlink won't fool php
+  cd buildenv
+  # This doesn't produce byte-for-byte identical output. We can get closer with --compress
+  HTTP_HOST=fakejethro.local php resources/less/jethro.less.php > lessoutput.css
+  bunx lessc --compress --include-path=resources/less lessoutput.css > ../resources/css/"jethro-${version}.css.new"
```
radarhere: https://lesscss.org/usage/ describes `--compress` as deprecated.
> This does an okay job but does not utilise all the tricks of dedicated css compression. In general, we recommend looking at third-party tools that clean and compress CSS after your Less has been transformed to CSS.

PR #1517, opened 2026-08-08
Fix MariaDB collations (#1511) and support emojis by standardizing on utf8mb4

### LINKED ISSUE #1511: Jethro breaks with 'Illegal mix of collations' on MariaDB 11.5.1+
Say we have a fresh Jethro installed in MariaDB 11.8.3 (Ubuntu 26.04 LTS), with the README-recommended setup:

```sql
	CREATE USER IF NOT EXISTS 'jethro'@'localhost' IDENTIFIED BY 'jethro';
	CREATE DATABASE jethro CHARACTER SET utf8 COLLATE utf8_unicode_ci;
	GRANT ALL PRIVILEGES ON jethro.* TO 'jethro'@'localhost';
	FLUSH PRIVILEGES;
```

After the setup screen, if we navigate to Rosters -> Define Roster Views, Jethro breaks with `General error: 1271 Illegal mix of collations for operation 'UNION'`:

Image

```php
Line 175 of File /home/jethro/code/current/include/jethrodb.php
	Array
(
    [0] => Array
        (
            [file] => /home/jethro/code/current/include/jethrodb.php
            [line] => 175
            [function] => execute
            [class] => PDOStatement
            [type] => ->
            [args] => Array
                (
                )

        )

    [1] => Array
        (
            [file] => /home/jethro/code/current/include/db_object.class.php
            [line] => 1245
            [function] => queryAll
            [class] => JethroDB
            [type] => ->
            [args] => Array
                (
                    [0] => SELECT roster_view.id, roster_view.name, roster_view.visibility, roster_view.show_on_run_sheet, members
				FROM (SELECT roster_view_id, GROUP_CONCAT(member_name ORDER BY order_num SEPARATOR ", ") as members
							FROM
							(
								(
									SELECT rvrm.roster_view_id as roster_view_id, CONCAT(COALESCE(rrc.name, ""), " ", rr.title) as member_name, rvrm.order_num as order_num
									FROM
									roster_view_role_membership rvrm
										JOIN (
											roster_role rr
											LEFT JOIN congregation rrc ON rr.congregationid = rrc.id
										) ON rvrm.roster_role_id = rr.id
								)
								UNION
								(
									SELECT rvdi.roster_view_id as roster_view_id, CONCAT(dic.name, " ", COALESCE(scc.category_name, rvdi.service_field)) as member_name, rvdi.order_num as order_num
									FROM
									roster_view_service_field rvdi
										JOIN congregation dic ON rvdi.congregationid = dic.id
										LEFT JOIN service_component_category scc ON ((rvdi.service_field LIKE "comps_%") AND (scc.id = SUBSTRING(rvdi.service_field, 7)))
								)
								ORDER BY order_num
							) res
							GROUP BY roster_view_id
						) m
						RIGHT OUTER JOIN roster_view on m.roster_view_id = roster_view.id
					ORDER BY roster_view.`name`
                    [1] => [Object/Array]
                    [2] => [Object/Array]
                    [3] => 1
                    [4] => 1
                )

        )

    [2] => Array
        (
            [file] => /home/jethro/code/current/include/db_object.class.php
            [line] => 1224
            [function] => _getInstancesData
            [class] => db_object
            [type] => ->
            [args] => Array
                (
                    [0] => [Object/Array]
                )

        )
...
```

jefft:
Fix MariaDB collations (#1511) and support emojis by standardizing on utf8mb4

Fixes #1511 (illegal mix of collations) and enables full Unicode, including emojis and characters outside the BMP, by
moving every Jethro database, table and column onto `utf8mb4 / utf8mb4_unicode_ci`. Supersedes the aborted PR #754.

# Two symptoms, one root cause

**Symptom 1 — #1511:** queries that UNION across tables blow up with

    General error: 1271 Illegal mix of collations for operation 'UNION'

**Symptom 2 — no emojis:** entering a 😢 anywhere in Jethro produces

    SQLSTATE[22007]: Invalid datetime format: 1366 Incorrect string value: '\xF0\x9F\x98\xA2'
    for column `jethro`.`_person`.`remarks` at row 1

Both are caused by tables ending up on inconsistent — and, for emojis, too-narrow — charsets. Jethro creates tables
through **two codepaths** and each picks up a different default:

1. `include/db_object.class.php` used to emit `... ENGINE=InnoDB DEFAULT CHARSET=utf8` (i.e. `utf8mb3`, no collation) —
   so the table got the server's default collation for `utf8mb3`, which varies by server version
   (`utf8mb3_general_ci` on MySQL 8 / MariaDB <11.5, `utf8mb3_uca1400_ai_ci` on MariaDB 11.5+).
2. Bespoke `CREATE TABLE`s in `db_objects/*` and `upgrades/*.sql` specify no charset and inherit the **database**
   default — which since 2018 we've told users to set to `utf8_unicode_ci` (i.e. `utf8mb3_unicode_ci`).

Result: a typical Jethro database mixes `utf8mb3_unicode_ci` (codepath 2) with `utf8mb3_uca1400_ai_ci` or
`utf8mb3_general_ci` (codepath 1). Cross-collation JOIN/UNION triggers #1511. And `utf8mb3` can't store any 4-byte code
point (emojis, some Japanese CJK like 吉, mathematical symbols like 𝛼) — hence symptom 2.

# Fix

Standardize on `utf8mb4 / utf8mb4_unicode_ci` everywhere:

- **New tables**: `include/db_object.class.php` no longer emits an explicit `DEFAULT CHARSET=utf8` — tables inherit the
  database default, which is now required to be `utf8mb4_unicode_ci`.
- **New installs**: `include/installer.class.php` refuses to install into a database whose default isn't
  `utf8mb4/utf8mb4_unicode_ci`, and prints the exact `CREATE DATABASE` command the operator should run.
- **PDO connection**: `include/jethrodb.php` uses `charset=utf8mb4` in the DSN (was `utf8`).
- **README**: updated to require `MariaDB 10.6+` or `MySQL 8+` and to show the required
  `CREATE DATABASE ... CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci` incantation.

# Existing installs: in-place repair

New file `upgrades/upgradefixes/2.40.0_fix_db_charset/dbcharsetutils.php` holds `DB_Charset_Utils::fix()`, which finds
every table whose collation isn't `utf8mb4_unicode_ci` and converts it. Three delivery paths, all idempotent:

- **CLI** — `php upgrades/2026-upgrade-to-2.40-utf8mb4.php` (run by the operator during upgrade). Reports per-table
  progress; if run on MySQL, where `ALTER DATABASE` needs `SUPER`, it prints the exact SQL a DBA should run and
  continues with everything else.
- **Admin dashboard** — `views/view_10_admin__8_upgrade.class.php` gains a "Database health" section showing per-table
  charset/collation and offering a manual re-run for post-upgrade confirmation.
- **Standalone script** — `scripts/database_health_check.php` produces the same report from the CLI (`--fix` to repair,
  `--detail` for per-column output).

If the operator forgets, `2026-upgrade-to-2.40.sql` sets a `NEEDS_UTF8MB4_UPGRADE` flag. The next home-page load
(`views/view_1_home.class.php`) then shows a notice to sysadmin users summarising the problem and linking to
**Admin → Upgrade**, where the fix can be reviewed and applied — it is not applied automatically.

# Corrupt-`latin1` edge case (PHP < 5.3.6)

PHP < 5.3.6 silently ignored `charset=utf8` in the PDO DSN, so Jethro installs that ever ran on such PHP stored raw
UTF-8 bytes inside `latin1` columns. Running a naïve `ALTER TABLE ... CONVERT TO CHARACTER SET utf8mb4` on those tables
would interpret each UTF-8 byte as a cp1252 character and re-encode, double-mangling the data
(`Cãlvín` → `CÃƒÂ£lvÃƒÂ­n`). This affects any install that once had:

- PHP < 5.3.6 (Jethro's stated floor until 2025-05-22, klkktkpn), AND
- MariaDB <11.6 or MySQL <8 (whose server default was `latin1`), AND
- a database created with just `create database jethro;` (inheriting the server default).

This problem scuttled our last attempt at switching to utf8mb4
(https://github.com/tbar0970/jethro-pmm/pull/754#issuecomment-1459052191).

`fix()` handles this with a **BLOB detour**:

1. For each `latin1` table, open a raw `charset=latin1` PDO connection and read every non-null text cell to validate
   the bytes with a strict UTF-8 decoder (`_isValidUTF8`).
2. If **every** row in **every** latin1 text column is valid UTF-8, treat the table as double-encoded and use the BLOB
   detour: `CONVERT TO CHARACTER SET binary` (strips the charset label, preserves bytes) → per-column
   `MODIFY ... CHARACTER SET utf8mb4` (retypes VARBINARY→VARCHAR etc.) → `DEFAULT CHARACTER SET utf8mb4`. The raw
   UTF-8 bytes are now interpreted as utf8mb4 and decode correctly. Any failure rolls back each column to its original
   latin1 type.
3. If any row contains a genuinely-cp1252 byte (e.g. `0x97` = em dash, which is not a valid UTF-8 lead byte), fall
   back to plain `CONVERT TO CHARACTER SET utf8mb4`, which correctly maps cp1252 bytes to their UTF-8 equivalents.
4. Legacy `ROW_FORMAT=COMPACT`/`REDUNDANT` tables are bumped to `DYNAMIC` in the same ALTER, because indexes that fit
   in 767 bytes as latin1 would exceed the limit as utf8mb4 (the `family` composite name/address index is 1368 bytes
   as utf8mb4).
5. Views need no rebuilding — their columns resolve to the base tables' charsets at query time. The fixer verifies the
   four Jethro views (`person`, `person_group`, `member`, `abstract_note`) still execute.

**Known limitation:** a table containing a mix of raw-UTF-8 and genuinely-cp1252 rows in the same column will be routed
to plain CONVERT and its UTF-8 rows will be double-encoded. This is inherent to any single-pass strategy.

# Testing

Playwright tests are deferred to a later commit.

# Files

```
    include/db_object.class.php                            # drop explicit DEFAULT CHARSET=utf8
    include/installer.class.php                            # refuse non-utf8mb4 DB, disable submit
    include/jethrodb.php                                   # DSN charset=utf8mb4
    upgrades/2026-upgrade-to-2.40.sql                      # set NEEDS_UTF8MB4_UPGRADE
    upgrades/2026-upgrade-to-2.40-utf8mb4.php              # CLI upgrader (idempotent)
    upgrades/upgradefixes/2.40.0_fix_db_charset/           # fixer + spec + test fixture
      dbcharsetutils.php                                   #   detection, BLOB detour, plain CONVERT
    upgrades/upgradefixes/2.40.0_fix_db_charset/           # snippet included in view_1_home.class.php
      utf8mb4_upgradecheck.php                             # 
    scripts/database_health_check.php                      # standalone CLI health report
    views/view_1_home.class.php                            # sysadmin notice linking to Admin → Upgrade
    views/view_10_admin__8_upgrade.class.php               # Admin → Upgrade → Database health
    README.md                                              # require utf8mb4, add MariaDB 10.6+
```

[jethro_utf8mb4_upgrade.webm](https://github.com/user-attachments/assets/bf7f1044-aea0-4937-a11f-a8225c5b4231)

### DIFF
--- a/upgrades/2026-upgrade-to-2.40-utf8mb4.php
+++ b/upgrades/2026-upgrade-to-2.40-utf8mb4.php
@@ -7,7 +7,7 @@
  * idempotent - safe to run more than once, and safe to run on a database that has
  * already been converted.
  *
- * Jethro 2.39 makes utf8mb4/utf8mb4_unicode_ci the standard character set for new
+ * Jethro 2.40 makes utf8mb4/utf8mb4_unicode_ci the standard character set for new
  * installs: the table-creation code (db_object.class.php, installer.class.php) and
  * the PDO connection charset now all use utf8mb4. This script brings existing
  * databases into line by converting every table not already on utf8mb4_unicode_ci
--- a/upgrades/2026-upgrade-to-2.40-utf8mb4.php
+++ b/upgrades/2026-upgrade-to-2.40-utf8mb4.php
@@ -1,0 +1,157 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Convert any remaining non-standard-collation tables to utf8mb4_unicode_ci.
+ *
+ * Run this script when upgrading from 2.38 (or earlier) to 2.39 or above. It is
+ * idempotent - safe to run more than once, and safe to run on a database that has
+ * already been converted.
+ *
+ * Jethro 2.39 makes utf8mb4/utf8mb4_unicode_ci the standard character set for new
+ * installs: the table-creation code (db_object.class.php, installer.class.php) and
+ * the PDO connection charset now all use utf8mb4. This script brings existing
+ * databases into line by converting every table not already on utf8mb4_unicode_ci
+ * (utf8mb3, latin1, or a non-standard utf8mb4 collation) to utf8mb4_unicode_ci.
+ *
+ * Notes:
+ * - The conversion is lossless for every character representable in utf8mb3 (the
+ *   Basic Multilingual Plane); 4-byte characters could not be stored in utf8mb3
+ *   tables, so there is nothing to lose.
+ * - Tables still on the legacy COMPACT/REDUNDANT row format are moved to DYNAMIC
+ *   in the same ALTER, because their indexes would otherwise exceed the 767-byte
+ *   InnoDB key limit once widened to utf8mb4 (eg the family table's composite
+ *   name/address index is 1368 bytes as utf8mb4). DYNAMIC is the modern default
+ *   row format anyway (3072-byte key limit).
+ * - Views need no rebuilding: their columns resolve to the base tables' charsets
+ *   at query time. The script verifies the four known views still execute.
+ * - Latin1 tables that contain raw UTF-8 bytes (from PHP < 5.3.6 era or
+ *   misconfigured connections) are handled safely via a BLOB detour:
+ *   CONVERT TO binary preserves bytes, then MODIFY cols to utf8mb4
+ *   reinterprets them without double-encoding. This matches the Admin
+ *   dashboard path (view_10_admin__8_upgrade.class.php → DB_Charset_Utils::fix()).
+ */
+
+if ((php_sapi_name() !== 'cli') && !defined('STDIN')) {
+	echo "This script must be run from the command line";
+	exit;
+}
+
+ini_set('display_errors', 1);
+define('JETHRO_ROOT', dirname(dirname($_SERVER['SCRIPT_FILENAME'])));
+set_include_path(get_include_path().PATH_SEPARATOR.JETHRO_ROOT);
+
+if (!is_readable(JETHRO_ROOT.'/conf.php')) {
+	echo "Jethro configuration file not found.  You need to copy conf.php.sample to conf.php and edit it before Jethro can run\n";
+	exit(1);
+}
+require_once JETHRO_ROOT.'/conf.php';
+define('DB_MODE', 'private');
+require_once JETHRO_ROOT.'/include/init.php';
+require_once JETHRO_ROOT.'/upgrades/upgradefixes/2.40.0_fix_db_charset/dbcharsetutils.php';
+/** @var JethroDB $db */
+$db = $GLOBALS['db'];
+echo "Converting tables in database ".ifdef('DB_DATABASE', 'unknown')." to utf8mb4_unicode_ci...\n";
+
+try {
+	$result = DB_Charset_Utils::fix();
+} catch (RuntimeException $e) {
+	// DB_Charset_Utils::fix() throws on MySQL when it cannot ALTER DATABASE
+	// (requires SUPER). The table conversion already completed; the
+	// database-default alignment is handled below with MySQL-aware logic.
+	$result = Array(
+		'converted' => Array(),
+		'errors' => Array(),
+		'database_default_aligned' => FALSE,
+	);
+}
+
+if (!empty($result['converted'])) {
+	foreach ($result['converted'] as $tbl) {
+		echo "  - ".$tbl."\n";
+	}
+	echo count($result['converted'])." table(s) converted to utf8mb4_unicode_ci.\n";
+}
+if (!empty($result['errors'])) {
+	echo "Errors:\n";
+	foreach ($result['errors'] as $error) {
+		echo "  * ".$error."\n";
+	}
+}
+
+// Align the database default charset as well, so tables created without an explicit
+// charset clause in the future (eg from later upgrades) also inherit utf8mb4.
+$dbname = $db->queryOne('SELECT DATABASE()');
+if (!empty($dbname)) {
+	// MySQL requires the SUPER privilege for ALTER DATABASE, which the
+	// application user typically lacks. On MySQL we check whether the
+	// database default even needs changing and either print the SQL to run
+	// or skip (if already correct). On MariaDB, ALTER DATABASE works
+	// with database-level privileges, so we attempt it.
+	$version = $db->queryOne('SELECT VERSION()');
+	$isMySQL = (stripos($version, 'MariaDB') === false);
+
+	if ($isMySQL) {
+		$current = $db->queryRow(
+			"SELECT DEFAULT_CHARACTER_SET_NAME AS charset, DEFAULT_COLLATION_NAME AS collation
+			 FROM information_schema.SCHEMATA WHERE SCHEMA_NAME = DATABASE()"
+		);
+		$needsAlignment = (empty($current) || $current['charset'] != 'utf8mb4' || $current['collation'] != 'utf8mb4_unicode_ci');
+		if ($needsAlignment) {
+			echo "\n";
+			echo "WARNING: Cannot set the database default charset/collation on MySQL\n";
+			echo "without the SUPER privilege. Please run the following command as a\n";
+			echo "MySQL superuser (e.g. root):\n";
+			echo "\n";
+			echo "  ALTER DATABASE `".$dbname."` CHARACTER SET utf8mb4 COLLATE utf8mb4_unicode_ci;\n";
+			echo "\n";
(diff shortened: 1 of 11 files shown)


tbar0970:
This all looks good, and it's been a long journey!

My one hesitation is about the "web (auto)" scenario. 
1) the db user which Jethro(web) is running as might not have sufficient permissions (SUPER) ?
2) there's no confirmation, it just happens
3) I believe it runs for even for the lowliest user (if they log in first), not just sys admins
4) there's a low-likelihood but high-impact risk that the latin1-converstion scenario will corrupt a lot of stuff.

All that being the case, I think it would be safer if the NEEDS_UTF8MB4_UPGRADE flag triggers a message on the home page, only shown to sysadmin users, directing them to the upgrade page to view and fix the DB health.

jefft:
> My one hesitation is about the "web (auto)" scenario.
> 
> the db user which Jethro(web) is running as might not have sufficient permissions (SUPER) ?

Yes, that would be the case for MySQL (no MariaDB) users, whose `jethro` db user isn't privileged.

> there's no confirmation, it just happens
> I believe it runs for even for the lowliest user (if they log in first), not just sys admins
> there's a low-likelihood but high-impact risk that the latin1-converstion scenario will corrupt a lot of stuff.
> All that being the case, I think it would be safer if the NEEDS_UTF8MB4_UPGRADE flag triggers a message on the home page, only shown to sysadmin users, directing them to the upgrade page to view and fix the DB health.

Yes, that is a more sensible approach! Thanks, I'll revise the patch.

jefft:
> All that being the case, I think it would be safer if the NEEDS_UTF8MB4_UPGRADE flag triggers a message on the home page, only shown to sysadmin users, directing them to the upgrade page to view and fix the DB health.

This change is now implemented, and the PR description revised with a video showing the warning.

There's also functional tests, but on a separate branch dependent on functional testing infrastructure:

https://github.com/tbar0970/jethro-pmm/actions/workflows/functests.yml?query=branch%3Apush-functests-utf8mb4

views/view_1_home.class.php
```diff
@@ -26,6 +26,7 @@ function printView()
 			require_once 'upgrades/upgradelibs/status_upgrader.class.php';
 			Status_Upgrader::runHTML();
 		}
+		require_once JETHRO_ROOT.'/upgrades/upgradefixes/2.40.0_fix_db_charset/utf8mb4_upgradecheck.php';
```
tbar0970: Can we check the NEEDS_UTF8MB4_UPGRADE flag before calling require_once (as in the other cases above). Otherwise we bear the cost of loading an extra file on every page load, even when it doesn't need to do anything.

upgrades/2026-upgrade-to-2.40-utf8mb4.php
```diff
@@ -0,0 +1,157 @@
+#!/usr/bin/env php
+<?php
+/**
+ * Convert any remaining non-standard-collation tables to utf8mb4_unicode_ci.
+ *
+ * Run this script when upgrading from 2.38 (or earlier) to 2.39 or above. It is
```
tbar0970:
```suggestion
 * Run this script when upgrading to 2.40 or above. It is
```

upgrades/2026-upgrade-to-2.40-utf8mb4.php
```diff
@@ -0,0 +1,157 @@
...
+ * idempotent - safe to run more than once, and safe to run on a database that has
+ * already been converted.
+ *
+ * Jethro 2.39 makes utf8mb4/utf8mb4_unicode_ci the standard character set for new
```
tbar0970:
```suggestion
 * Jethro 2.40 makes utf8mb4/utf8mb4_unicode_ci the standard character set for new
```

views/view_1_home.class.php
jefft: Good point! Fixed.

jefft:
Changes applied.

PR #1520, opened 2026-08-08
Add Note: convert from full page view to modal dialog with AJAX submission

### LINKED ISSUE #1549: Make Add Note to sequence of people faster by using modal, not window popup
Jethro sometimes shows a list of people, with 'Add Note' links for each.

Image

This happens:
 - on Persons -> List All
 - on Person Report results, if `A link to add a note` column is selected
 - On Attendance -> Display

After adding a note to one person, typically one would want to proceed through the list, perhaps to add more notes. Jethro isn't great for this:
 - In Persons -> List All, clicking 'Add Note' redirects you to a 'Add Note to ...' page. After filling in the note, one is redirected to that person's view page, not the person list.
 - In Attendance Display, clicking 'Add Note' opens a new popup window. After adding the note, you're left viewing that person. You have to close the popup yourself. This is fine on desktop, but on mobile there's little indication your original attendance marking window is even still open.
 - Person Report results handle this a little better: we again get a popup window, but this time the window closes itself once the note is entered, returning us to the reloaded report list (the reload loses checkbox state).

[jethro_addnote_old.webm](https://github.com/user-attachments/assets/78ebb2e8-43e3-4cf4-ab5e-b2d536f66de8)

I suggest popup windows are not the right tool for 'Add Note'. Jethro should use a modal popup instead:

Image

This removes the need for window management.

jefft:
This allows one to rapidly add Notes to a sequence of people.

Old way: each click takes you to the 'Add note' page, and then you need to re-do your search:

[jethro_addnote_old.webm](https://github.com/user-attachments/assets/78ebb2e8-43e3-4cf4-ab5e-b2d536f66de8)

New way: 'Add Note' links now open a Bootstrap modal and submit via AJAX to a new ?call=note endpoint instead of navigating away to ?view=_add_note_to_person.

[jethro_addnote_20260901.webm](https://github.com/user-attachments/assets/cd40b62a-adfd-43da-840d-bcc92f113ae0)

Changes: 

- include/general.php: print_note_modal_once() — static-guarded modal
  printer, replacing the ad-hoc guard idioms previously duplicated in
  view_person, person_list and person_query.
- person_query.class.php: the report note_link column renders a modal
  trigger (bulk name lookup for data-name); reports showing a Notes
  column set data-refresh so the page reloads after saving (successor of
  the old then=refresh_opener popup behaviour). The modal is printed once
  per page after the bulk-action form; the include flag is captured before
  the include because bulk_actions.template.php reassigns the local
  $params in printResults scope.
- jethro.js: note-modal open/submit handlers support data-refresh
  (location.reload() ~1s after a successful save).
- view_person.template.php: the Add Note link still used the old full-page
  URL while the modal was printed but untriggered; it is now a real
  trigger.
- person_list.template.php: the modal was nested inside the bulk-action
  form, where its subject/details inputs would be submitted with bulk
  actions; moved outside the form.
- attendance_record_set.class.php, view_6_attendance__1_record.class.php,
  view_6_attendance__2_display.class.php: roll/record/display Add Note
  links converted; rows gained data-personid for the note-success
  highlight; the modal is printed once per view.
- view_family.template.php: single-member "Add Person Note" converted.
- tb_lib.js: stop the .clickable-rows js following the link if a modal is
  attached.

Middle-clicking 'Add Note' (opening a new tab) still works, i.e. the link
still points to the 'Add Note' page. A left-click click is just intercepted
by js.

Still on the old flow: the bulk multi-person "Add Note" action (needs
multi-recipient support in the modal JS; the endpoint already accepts
personid arrays) and family notes (_add_note_to_family — a separate
Family_Note feature).

Fixes #1549


tbar0970:
Functionality looks good.  But for code consistency can we please:
* rename the call to call_add_note
* change note_modal.php into a template in the templates dir

Does it work OK on mobile?

jefft:
> rename the call to call_add_note
> change note_modal.php into a template in the templates dir

Sure, done.

> Does it work OK on mobile?

It's just as well you're reviewing! Mobile was broken:

[jethro_note_modal_broken.webm](https://github.com/user-attachments/assets/6459fa6f-dc1c-48c2-aa30-48622eb41b39)

The bootstrap `modal` class sets a fixed 560px width for big screens:

https://github.com/tbar0970/jethro-pmm/blob/83b2ee4fc9f69d2fcf14c7cfb887ec3b210b9fd8/resources/less/modals.less#L24-L29

but then switches to `auto` width on smaller screens, and changes `left` from `50%` to a fixed `20px`:

https://github.com/tbar0970/jethro-pmm/blob/83b2ee4fc9f69d2fcf14c7cfb887ec3b210b9fd8/resources/less/responsive-767px-max.less#L103-L112

My CSS widened 560px to 700px regardless of viewport, and interacted particularly badly with the `left: 20px`:

```css
.note-modal {
	width: 700px;
	margin-left: -350px;
 }
```

I have fixed the new CSS to set width to just under 100% of the viewport width, up to a maximum 700px, and let the CSS algorithm center it (in all viewport sizes).

```css
.note-modal {
	width: calc(100vw - 2rem);
	max-width: 700px;
	left: 0;
	right: 0;
	margin-left: auto;
	margin-right: auto;
}
```

Also fixed a problem where the button text disappeared in narrow viewports.

[jethro_note_modal_fixed.webm](https://github.com/user-attachments/assets/d2072df7-1473-48ad-8a15-9475b0cad92f)

tbar0970:
One more thing I wonder about:
How does this go in person reports?
Reports may include a "current notes" column.
Currently, after you add a note via the add-note popup window, the report page refreshes so that "current notes" is up to date.
It'd be good to check that still works with the new approach...

jefft:
> How does this go in person reports?
> Reports may include a "current notes" column.
> Currently, after you add a note via the add-note popup window, the report page refreshes so that "current notes" is up to date.
> It'd be good to check that still works with the new approach...

Hm, so far this PR only converts Persons -> List All to use a modal. The very similar 'Add Note' from report results still uses a popup, successfully:

https://github.com/user-attachments/assets/6f5441e1-29a2-4e5c-a521-f9535bb266db

It is inconsistent having modals here and popups there. I'll create a second PR to convert the other 'Add Note' occurrences to use modals instead of pop-ups.

jefft:
PR updated to cover 'List All', report and attendance use-cases, per newly created #1549

New video:

[jethro_addnote_20260901.webm](https://github.com/user-attachments/assets/cd40b62a-adfd-43da-840d-bcc92f113ae0)

Also, I fixed the implementation so that middle-clicking (vs. left clicking) 'Add Note' opens the original 'Add Note' page. The 'Add Note' URL still points to the original Add Note page, and it's only Javascript that makes it a modal.

sevasek:

PR #1530, opened 2026-08-26
Fix utf8mb4 fixer if DB_DSN is used

jefft:
Follow-up fix for #1511, for users specifying `DB_DSN` instead of `DB_HOST/DB_PORT` etc.


sevasek:

PR #1543, opened 2026-08-29
Right-align service schedule date-row dropdown menus (#1542)

### LINKED ISSUE #1542: Drop-down in Service Schedule is clipped to the right of viewport
To replicate:
 - Go to Services -> List All
 - Select a date range showing services
 - Narrow the window slightly
 - Click the down arrow at the end of a service row

The drop-down appears off to the right of the viewport, requiring horizontal scrolling to view:
Image

Replicated in 2.38.0 and 2.39.0

jefft:
The per-date "generate service documents" dropdown in services__list_all sits in the right-most column of the table, and its menu opened with left:0 from the toggle - running off the right edge of narrow viewports. Add pull-right (Bootstrap 2's right-aligned dropdown-menu modifier, as already used by the service-plan tools dropdown) so the menu opens leftward and stays on screen at any width.

Before the fix:

image

With fix applied:

image

Fixes #1542

### DIFF
--- a/views/view_8_services__1_list_all.class.php
+++ b/views/view_8_services__1_list_all.class.php
@@ -399,7 +399,7 @@ class View_Services__List_All extends View
 						<td>
 							<span class="dropdown">
 							<a class="dropdown-toggle" data-toggle="dropdown" href="#"><i class="icon-chevron-down"></i></a>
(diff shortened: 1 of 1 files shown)


sevasek:

PR #1545, opened 2026-08-30
Handle requests arriving without Referer/User-Agent headers (#1544)

### LINKED ISSUE #1544: Undefined array key "HTTP_REFERER" if runsheet 'Edit Personnel' link is copy and pasted
When displaying a service, there may be a 'Edit Personnel' link (if any roster views have 'Show On Run Sheet' set). In normal use, clicking 'Edit Personnel' takes one to the Edit Roster Assignments page for the service's date/service. Clicking 'Save' saves the personnel change and then redirects one back to the service sheet.

However, if one right-clicks 'Edit Personnel', copies the link and pastes it into a new tab, Jethro breaks: 

```
Undefined array key "HTTP_REFERER"
Line 22 of /home/jethro/code/current/views/view_7_rosters__2_edit_roster_assignments.class.php

USER:       194
REFERER:   
USER_AGENT: Mozilla/5.0 (Linux; Android 10; K) AppleWebKit/537.36 (KHTML, like Gecko) Chrome/[151.0.0.0](http://151.0.0.0/) Mobile Safari/537.36

REQUEST_ID: 0a057c51836e2c574033e7a1618b6120

REQUEST:
Array
(
    [view] => rosters__edit_roster_assignments
    [viewid] => 4
    [start_date] => 2026-08-30
    [end_date] => 2026-08-30
    [goback] => 1
)
```

https://github.com/user-attachments/assets/895847ec-ef82-4b03-8668-1db905f51765

Copying and pasting the link means the HTTP_REFERER header is not set, and this header is relied upon for the redirect. If HTTP_REFERER is not set, Jethro should simply save personnel edits and not redirect.

jefft:
Handle requests arriving without Referer/User-Agent headers (#1544)

In a few places Jethro assumed a valid Referer or User-Agent header. This was a
bad assumption: browsers are free to omit the Referer and User-Agent headers
(direct URL entry, bookmarks, tab restore, privacy stripping, HTTPS->HTTP
downgrade), and PHP 8 turns the resulting "Undefined array key" warnings into
system error reports. redirect(-1) was worse: it sent a "Location: " header with
an empty URL. 

This commit fixes all places we assumed a header. We now use array_get() and
provide sensible fallbacks that preserve existing behaviour when the headers are
present:

- view_7_rosters__2_edit_roster_assignments: leave roster_backto unset, so
  saving falls back to the normal roster display redirect (#1544)
- redirect(-1): fall back to the home view, matching system_controller's
  array_get($_REQUEST, 'view', 'home') default
- print_widget() int fields: no UA means type="number" (not "tel")
- get_email_href(): no UA means the Chrome-on-mac mailto tweak is skipped
- person mobile_tel: no UA omits the iMessage link; size_detector
  debug output shows an empty UA

Fixes #1544

### DIFF
--- a/db_objects/person.class.php
+++ b/db_objects/person.class.php
@@ -295,7 +295,7 @@ class Person extends DB_Object
 						$printedModal = TRUE;
 					}
 				}
-				if (FALSE !== strpos($_SERVER['HTTP_USER_AGENT'], 'Macintosh')) {
+				if (FALSE !== strpos(array_get($_SERVER, 'HTTP_USER_AGENT', ''), 'Macintosh')) {
 					// on mac we can use the messages app
 					$msg = _('SMS via iMessage');
 					$links[] = '<a href="imessage:'.ents($value).'"><i class="icon-envelope"></i> '.$msg.'</a>';
--- a/include/general.php
+++ b/include/general.php
@@ -160,7 +160,11 @@ function redirect($view, $params=Array(), $hash='')
 	session_write_close();
 	if ($view == -1) {
 		// go back
-		$url = $_SERVER['HTTP_REFERER'];
+		$url = array_get($_SERVER, 'HTTP_REFERER');
+		if (!$url) {
+			// No referer to go back to (direct entry, privacy stripping etc) - go home instead
+			$url = build_url(Array('view' => 'home'));
+		}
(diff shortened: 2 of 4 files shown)


sevasek:

PR #1548, opened 2026-08-31
Fix unexpanded roster role link on the personnel format help page (#1547)

### LINKED ISSUE #1547: Broken links in service component personnel help
To replicate:
 - Click Services -> Component Library -> Add
 - Locate the 'Personnel' field
 - click 'help and examples'. This takes one to /?call=service_comp_help_personnel_format
 - Scroll to the the help's 'roster role keywords' section, where each roster role is hyperlinked

The role hyperlinks have untranslated code in it, with links like `/?view=rosters__define_roster_roles&roster_roleid=${titleinfo[%27id%27]}`:

Image

Clicking this link results in stacktraces beginning with:

```
array_map(): Argument #2 ($array) must be of type array, null given
Line 112 of /home/jethro/code/current/db_objects/roster_role.class.php
```

Replicated in 2.38.0 and 2.39.0.

jefft:
The "service personnel value" link on the Personnel Format help page
emitted a raw-HTML href containing ${titleinfo['id']}, which PHP never
interpolates because the link sits in template HTML rather than a PHP
string. Clicking it navigated to rosters__define_roster_roles with the
literal ${titleinfo['id']} text as roster_roleid; the (int) cast turned
that into 0, constructing an unloaded Roster_Role whose summary
rendering passed null to array_map(), causing a TypeError.

Fixes #1547

image

### DIFF
--- a/db_objects/roster_role.class.php
+++ b/db_objects/roster_role.class.php
@@ -112,7 +112,7 @@ class Roster_Role extends db_object
 				echo implode(', ', array_map(function ($val) {
 					$group = new Person_group($val);
 					return '<a href="?view=groups&groupid='.(int)$val.'">'.ents($group->getValue('name')).'</a>';
-				}, $this->getValue($name)));
+				}, $this->getValue($name) ?? []));
 				break;
 			case 'volunteer_group':
 				if ($val = $this->getValue($name)) {
--- a/calls/call_service_comp_help_personnel_format.class.php
+++ b/calls/call_service_comp_help_personnel_format.class.php
@@ -80,7 +80,7 @@ class call_service_comp_help_personnel_format extends Call
                             <code>%<?php echo $titleinfo['title_uppercase']; ?>_n_FIRSTNAME%</code><br>
                         </td>
                         <td>
-                            <a href="?view=rosters__define_roster_roles&roster_roleid=${titleinfo['id']}"><?php echo $titleinfo['title']; ?></a>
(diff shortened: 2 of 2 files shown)


db_objects/roster_role.class.php
```diff
@@ -112,7 +112,7 @@ function printFieldValue($name, $value=NULL)
 				echo implode(', ', array_map(function ($val) {
 					$group = new Person_group($val);
 					return '<a href="?view=groups&groupid='.(int)$val.'">'.ents($group->getValue('name')).'</a>';
-				}, $this->getValue($name)));
+				}, $this->getValue($name) ?? []));
```
radarhere, line 115:
```suggestion
				}, $this->getValue($name) ?: []));
```
I'm used to `??` being used to check if something is set, which doesn't quite apply here.

jefft: > I'm used to ?? being used to check if something is set, which doesn't quite apply here.

Doesn't it? We want a fallback array if `$this->getValue($name)` is not set, which is what `?? []` does.

`?:` tests for truthiness, a broader class which covers null but also `""`, `0` and `false`. `$this->getValue($name)` should never be `""` or `0` or `false`, and if it is, we wouldn't want that bug being masked here.

radarhere: Ok, if you're thinking about general truthiness, then no worries.

sevasek:

PR #1558, opened 2026-09-15
Fix service run-sheet wipe when saving items fails (#1557)

### LINKED ISSUE #1557: Service runsheet: 'Copy from previous' wipes runsheet is Ad-hoc item present
To replicate:
 - Services -> List All
 - Find a slot with an existing service and click the icon
 - Click 'Edit'
 - Add an ad-hoc item to the runsheet, by clicking the drop-down on any runsheet item and selecting 'Add ad-hoc item below'
 - Save the runsheet
 - Click 'Copy from previous'
 - Click 'Copy items'

Jethro breaks with:

```php
SQLSTATE[23000]: Integrity constraint violation: 1048 Column 'personnel' cannot be null
Line 707 of /home/jethro/code/current/db_objects/service.class.php

REQUEST:
Array
(
    [date] => 2026-09-20
    [congregationid] => 29
    [view] => services
    [editing] => 1
    [copy_service_id] => 188
    [copy_category_ids] => Array
        (
            [0] => 1
            [1] => 2
            [2] => 3
            [3] => 4
            [4] => !
        )

    [action] => copy
)


BACKTRACE:
Array
(
    [0] => Array
        (
            [file] => /home/jethro/code/current/db_objects/service.class.php
            [line] => 707
            [function] => exec
            [class] => PDO
            [type] => ->
        )

    [1] => Array
        (
            [file] => /home/jethro/code/current/views/view_8_services.class.php
            [line] => 73
            [function] => saveItems
            [class] => service
            [type] => ->
        )

    [2] => Array
        (
            [file] => /home/jethro/code/current/include/system_controller.class.php
            [line] => 145
            [function] => processView
            [class] => View_services
            [type] => ->
        )


```

https://github.com/user-attachments/assets/253e53a8-2a39-477a-b024-0692408c990b

jefft:
Copying a service whose run sheet contains an ad-hoc item with empty personnel crashed with "Column 'personnel' cannot be null" and, worse, wiped the target service's run sheet, because saveItems() deleted all existing items (autocommit) before an INSERT that then failed.

Three-layer fix:
- Service::getItems() resolves personnel via an IF() that falls through to scc.personnel_default, which is NULL for ad-hoc items (componentid IS NULL) because both LEFT JOINs miss. Wrap the expression in IFNULL(..., '') so the getter honours service_item.personnel's NOT NULL contract. '' remains the established "inherit from component" sentinel.
- Service::saveItems() coerces personnel to string with (string)array_get(...) before quoting, guarding every write path (e.g. a POST missing a personnel field) against emitting literal SQL NULL.
- Service::saveItems() now runs the DELETE and the multi-row INSERT inside one transaction (beginTransaction/commit, rollback + rethrow on Throwable), so any future INSERT failure leaves the previous items intact instead of destroying the run sheet.

Fixes #1557

[jethro_copyprevious_adhoc_fixed.webm](https://github.com/user-attachments/assets/2bce38b7-e0dd-4423-96e6-aada5d7e5995)


db_objects/service.class.php
```diff
@@ -736,7 +767,7 @@ public function getItems($withContent=FALSE, $ofCategoryID=NULL)
 					'.($withContent ? 'sc.content_html, sc.credits, ' : '').'
 					IFNULL(IF(LENGTH(sc.runsheet_title_format) = 0, scc.runsheet_title_format, sc.runsheet_title_format), "%title%") AS runsheet_title_format,
 					IFNULL(IF(LENGTH(sc.handout_title_format) = 0, scc.handout_title_format, sc.handout_title_format), "%title%") AS handout_title_format,
-					IF(LENGTH(si.personnel) > 0, si.personnel, IF(LENGTH(sc.personnel) > 0, sc.personnel, scc.personnel_default)) as personnel,
+				IFNULL(IF(LENGTH(si.personnel) > 0, si.personnel, IF(LENGTH(sc.personnel) > 0, sc.personnel, scc.personnel_default)), "") as personnel,
```
radarhere, line 770:
```suggestion
					IFNULL(IF(LENGTH(si.personnel) > 0, si.personnel, IF(LENGTH(sc.personnel) > 0, sc.personnel, scc.personnel_default)), "") as personnel,
```
Currently the indentation doesn't line up.
https://github.com/tbar0970/jethro-pmm/blob/4ca60935382c0b1cb9779c6eb21e2aa876b401c3/db_objects/service.class.php#L768-L772

tbar0970:
Happy enough with the solution, especially that the DELETE and INSERTs should all be in the one transaction (not sure why they weren't already).

But I'm intrigued how the NULL value was getting in there... when the run sheet is saved, every row includes a personnel field , so how is it becoming NULL rather than a blank string?

sevasek:

PR #1559, opened 2026-09-15
Hide "View Roster Role Descriptions" in members area if not allowed

radarhere:
In the Members area, let's not show
https://github.com/tbar0970/jethro-pmm/blob/d222048ca0322f88a0a35325663f11e80d89faed/members/views/view_3_rosters.class.php#L50
if the link is just going to take the user to
https://github.com/tbar0970/jethro-pmm/blob/d222048ca0322f88a0a35325663f11e80d89faed/public/views/view_0_roster_role_description.class.php#L32-L36


sevasek:

PR #1561, opened 2026-09-16
Roster iCal URL should be absolute when BASE_URL is inferred (#1560)

### LINKED ISSUE #1560: iCal URL incorrectly relative if BASE_URL is inferred
To replicate:
 - Do _not_ specify `BASE_URL` in Jethro's `conf.php` (if unspecified it will be inferred from the request).
 - Log in to the members interface as anyone
 - Click the 'Subscribe' link alongside 'My Roster'
 - Click 'enable iCal feed' if necessary

The generated roster iCal Feed is relative, when it should be absolute:

Image

jefft:
Jethro rendered the roster iCal URL as relative if BASE_URL wasn't explicitly defined. It should always be absolute.

Fixes #1560


sevasek:

PR #1562, opened 2026-09-16
Allow members to re-assign their rostered roles

radarhere:
In the 'Members area', users can see which roles they are rostered on for.

Screenshot 2026-09-16 at 10 13 32 pm<br />

See what you think of this idea - if the event of someone wanting to swap with a different pers on a roster, perhaps it would be helpful for them to be able to make that update themselves?

I have imagined a 'Swap' link

Screenshot 2026-09-16 at 10 13 24 pm<br />

which, if there is a volunteer group, can open to a dropdown to select the other individuals in the group

Screenshot 2026-09-16 at 10 08 34 pm

or if there is no volunteer group, can open to an autocomplete field.

Screenshot 2026-09-16 at 10 08 25 pm

I understand if you would like to not put this kind of control in the 'Members area'. It does presume that the user has separately confirmed with the new person that this is acceptable.


jefft:
> It does presume that the user has separately confirmed with the new person that this is acceptable.

If I saw a 'Swap' link, I would imagine Jethro is offering to _coordinate_ a swap, i.e. find a replacement by contacting other non-rostered volunteers. That would be a genuinely useful feature, but hard to implement.

This feature would be better labelled as 'Record agreed swap'. So people realise there is no magical roster fairy arranging swaps for them in the background.

I'm still not sure I'd want to allow anyone to modify rosters like this, especially as Jethro has no roster history or audit trail.

radarhere:
Happy to update the link text to whatever you like.

We could conceivably show both the original and the new person, so that a roster would say 'Jane Doe (swapped with John Smith)', making this a non-destructive operation that clearly identifies that John Smith said that Jane Doe would step in.

But yes, I understand this may or may not be desirable.

tbar0970:
I think the ideal would be a tool to *request* a swap.
Joe Bloggs can't come so he clicks a few buttons and Jane Doe gets an email saying "Joe Bloggs has asked you to cover Bible Reading on 20 Sept.  To accept, click here".

One step beyond this would be "Joe Bloggs has asked you to do Bible Reading on 20 Sept, and offered to cover your slot on 27 Sept. To accept, click here".

Probably ought to be configurable so churches can disable it if they don't like it.

sevasek:

radarhere:
> I think the ideal would be a tool to _request_ a swap. Joe Bloggs can't come so he clicks a few buttons and Jane Doe gets an email saying "Joe Bloggs has asked you to cover Bible Reading on 20 Sept. To accept, click here".
 
> Probably ought to be configurable so churches can disable it if they don't like it.

I've pushed commits for these items.

tbar0970:
Good progress. Some feedback:
1) it's doing too much in the members_home view.  You should add members/views/process_roster_swap.class.php.  And SQL operations should go in a db_object class (probably roster_role_assignment).
2) it's a bit too trusting. As it stands, anybody could construct a URL to "accept a swap" and change the roster however they want. We need to either record the swap request in the database, or somehow sign the acceptance URL so they can't be just made up ad hoc.

radarhere:
2. I had thought of this, but the vulnerability would only allow them to reassign the role to themselves. If John Smith wants to take over every role, then ok? Everyone at least knows who the culprit is. But sure, I'll put together a change to be more traditionally secure.

radarhere:
1. Ok, I've moved the code around.
2. I've added a pending_personid column to hold the person that was asked, and when the acceptance link is clicked, this is checked and then cleared.

